PR #4, opened 2021-06-07
create subscription for the storage system

iamniting:
As part of the PR, the operator will create a subscription based on the given kind in storagesystem cr.


controllers/conditions.go
```diff
@@ -0,0 +1,73 @@
...
+func (r *StorageSystemReconciler) setConditionResourcePresent(instance *odfv1alpha1.StorageSystem, logger logr.Logger) (bool, error) {
+
+	var requeue bool
+	var err error
+
+	if instance.Spec.Kind == odfv1alpha1.StorageCluster {
+		logger.Info("get storageCluster")
+		// TODO add types for ocs
+		//storageCluster := &ocsv1.StorageCluster{}
+		//err = r.Client.Get(context.TODO(), types.NamespacedName{Name: instance.Spec.Name, Namespace: instance.Spec.NameSpace}, storageCluster)
```
jarrpa, line 50: We shouldn't need this since we want to rely on OLM to handle this dependency.

iamniting: Still, we have to check the status if storageCluster present then set the condition accordingly right

jarrpa (approved):

PR #7, opened 2021-06-10
Add controller unit tests

iamniting:

### DIFF
--- a/controllers/subscriptions_test.go
+++ b/controllers/subscriptions_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2021 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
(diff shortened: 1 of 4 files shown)


controllers/subscriptions_test.go
```diff
@@ -0,0 +1,84 @@
...
+		existingSubscriptions := &operatorv1alpha1.SubscriptionList{}
+		err = fakeReconciler.Client.List(context.TODO(), existingSubscriptions)
+		assert.NoError(t, err)
+
+		if !tc.expectedSubscription {
+			assert.Equal(t, 0, len(existingSubscriptions.Items))
+		} else {
+			assert.Equal(t, 1, len(existingSubscriptions.Items))
+			assert.Equal(t, "ibm-operator", existingSubscriptions.Items[0].Spec.Package)
+			assert.Equal(t, "alpha", existingSubscriptions.Items[0].Spec.Channel)
```
jarrpa, line 81: This is fine for now, but I would like to see this be a little less hardcoded. Add a comment like `//TODO: Parameterize this`.

jarrpa (changes requested):
Overall looks good. I would request that you add labels to the test cases so we can get descriptive messages of what each case is actually testing. You can see an example here: https://github.com/openshift/ocs-operator/blob/master/controllers/persistentvolume/persistentvolume_controller_test.go (note specifically the `Logf()` in the test loop.

controllers/subscriptions_test.go
iamniting: Done

iamniting:
> Overall looks good. I would request that you add labels to the test cases so we can get descriptive messages of what each case is actually testing. You can see an example here: https://github.com/openshift/ocs-operator/blob/master/controllers/persistentvolume/persistentvolume_controller_test.go (note specifically the `Logf()` in the test loop.

Done

jarrpa (approved):
LGTM

PR #9, opened 2021-06-15
api: Change StorageSystem spec

iamniting:
* Add kubebuilder tags which will validate CR upon create/update operation
* Append the apiVersion in the names of StorageKind variable

### DIFF
--- a/api/v1alpha1/storagesystem_types.go
+++ b/api/v1alpha1/storagesystem_types.go
@@ -24,17 +24,17 @@ import (
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
-//+kubebuilder:validation:Enum=FlashSystemCluster;StorageCluster
+//+kubebuilder:validation:Enum=flashsystemcluster.odf.ibm.com/v1alpha1;storagecluster.ocs.openshift.io/v1
 
 // StorageVendor captures the type of storage vendor
 type StorageKind string
 
 const (
 	// FlashSystemCluster represents the ibm flashsystem
-	FlashSystemCluster StorageKind = "FlashSystemCluster"
+	FlashSystemCluster StorageKind = "flashsystemcluster.odf.ibm.com/v1alpha1"
 
 	// StorageCluster represents the openshift container storage
-	StorageCluster StorageKind = "StorageCluster"
+	StorageCluster StorageKind = "storagecluster.ocs.openshift.io/v1"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


api/v1alpha1/storagesystem_types.go
```diff
@@ -51,12 +51,13 @@ type StorageSystemSpec struct {
 	// Name describes the name of managed storage vendor CR
 	Name string `json:"name,omitempty"`
 
-	//+kubebuilder:validation:Optional
+	//+kubebuilder:validation:Required
 
 	// NameSpace describes the namespace of managed storage vendor CR
 	NameSpace string `json:"nameSpace,omitempty"`
```
jarrpa, line 57: If we're saying this field is required, then I would think we shouldn't have `omitempty`? Similar for `Name` above?

jarrpa (changes requested):
"updation" is not a word. ;) Please replace with "update" in the commit message.

And just to be picky, can you structure the comments like this? https://github.com/openshift/ocs-operator/blob/master/api/v1/storagecluster_types.go#L140-L143 It would make it clearer what field they're modifying.

api/v1alpha1/storagesystem_types.go
iamniting: This is mainly done to make API's backward compatible. As we are adding `Required` tag no one will be able to create a CR without the values. But in the code, we can skip those values if required. Let me know what do you think.

iamniting:
> "updation" is not a word. ;) Please replace with "update" in the commit message.
> 
> And just to be picky, can you structure the comments like this? https://github.com/openshift/ocs-operator/blob/master/api/v1/storagecluster_types.go#L140-L143 It would make it clearer what field they're modifying.

Done

api/v1alpha1/storagesystem_types.go
jarrpa: Huh... backwards compatible with what? Does this affect the API code generated in such a way that it would impact how we can interact with it?

iamniting: The first answer from this post should help. However, it is a protobuf related post but the problem I want to explain is the same.

https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3

jarrpa: The entire post is talking about why required was *removed* and why it's *bad*. Why are you adding it here?

iamniting: So with this post what I want to show is that protobuf has removed this functionality and now generate all go code with the `omitempty` and I also want to follow the same thing.

jarrpa: None of that is clear, `omitempty` is not mentioned anywhere in that post. I still don't understand how this is relevant.

iamniting: Let's discuss this in some of our mtg. I will remove omitempty. If we feel to add we can always do it.

jarrpa (approved):
LGTM

jarrpa (approved):

PR #16, opened 2021-06-25
Create StorageCluster controller

iamniting:
* Create StorageCluster controller scaffold
* Implement storagecluster controller

The controller will create an storage system CR if there is not one
present already for the storagecluster.

This will help in upgrade cases from 4.8 to 4.9


iamniting:
@jarrpa We need to release the storageCluster v1 API so that I can import that and make use of it in the controller.

controllers/storagecluster_controller.go
```diff
@@ -19,10 +19,22 @@ package controllers
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	ocsv1 "github.com/openshift/ocs-operator/api/v1"
+	odfv1alpha1 "github.com/red-hat-data-services/odf-operator/api/v1alpha1"
+)
+
+const (
+	HasStorageSystem = "odf.openshift.io/hasStorageSystem"
```
jarrpa, line 37: Annotations should not have capital letters, best I can tell. Also OpenShift has somewhat weird conventions on how they generally want annotations, and really really don't want booleans. Maybe change this to something like `storagesystem.odf.openshift.io/watched-by` and set it's value to the corresponding StorageSystem name.

controllers/storagecluster_controller.go
```diff
@@ -45,17 +58,111 @@ type StorageClusterReconciler struct {
...
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
 func (r *StorageClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+
+	instance := &ocsv1.StorageCluster{}
+	err := r.Client.Get(context.TODO(), req.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("storagecluster instance not found")
```
jarrpa, line 67: Make sure all instances of `StorageCluster` and `StorageSystem` have proper camelcase.

controllers/storagecluster_controller.go
```diff
@@ -45,17 +58,111 @@ type StorageClusterReconciler struct {
...
+		return ctrl.Result{}, err
+	}
+
+	logger.Info("storagecluster instance found")
+
+	// get list of StorageSystems
+	storageSystemList := &odfv1alpha1.StorageSystemList{}
+	err = r.Client.List(context.TODO(), storageSystemList, &client.ListOptions{Namespace: instance.Namespace})
+	if err != nil {
+		logger.Error(err, "failed to list the storageSystem")
```
jarrpa, line 80: We should start trying to adhere to better logging standards. Make sure all messages start with a capital letter and end with a period.

controllers/storagecluster_controller.go
```diff
@@ -45,17 +58,111 @@ type StorageClusterReconciler struct {
...
+		return ctrl.Result{}, err
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func filterStorageSystem(storageSystemList *odfv1alpha1.StorageSystemList, storageClusterName string) *odfv1alpha1.StorageSystem {
+
+	for _, ss := range storageSystemList.Items {
+		if ss.Spec.Name == storageClusterName && ss.Spec.Kind == odfv1alpha1.StorageCluster {
```
jarrpa, line 125: Should `odfv1alpha1.StorageCluster` be `ocsv1.StorageCluster`?

controllers/conditions.go
```diff
@@ -56,15 +56,15 @@ func (r *StorageSystemReconciler) setConditionResourcePresent(instance *odfv1alp
 	}
 
 	if err == nil {
-		logger.Info("set condition", odfv1alpha1.ConditionResourcePresent, corev1.ConditionTrue)
+		logger.Info("set condition", string(odfv1alpha1.ConditionResourcePresent), string(corev1.ConditionTrue))
```
jarrpa, line 59: This seems unrelated to this PR. Is this fixing an issue that was introduced before?

jarrpa (changes requested):

controllers/conditions.go
iamniting: yes

controllers/storagecluster_controller.go
iamniting: no `odfv1alpha1.StorageCluster` is just a string and not the whole type like `ocsv1.StorageCluster`

https://github.com/red-hat-data-services/odf-operator/blob/main/api/v1alpha1/storagesystem_types.go#L37

controllers/storagecluster_controller.go
iamniting: Done

controllers/storagecluster_controller.go
iamniting: Done

controllers/storagecluster_controller.go
iamniting: Done

controllers/storagecluster_controller.go
jarrpa: ....why are we defining the StorageCluster Kind as a string in odf-operator? This seems backwards. Shouldn't we be referencing or generating it from somewhere in `ocsv1`?

iamniting: We can do that, But I want to keep control of these variables in my own hands So that we can keep them the way we want instead of asking the vendors to do that.

for example it was `StorageCluster StorageKind = "StorageCluster"` and we changed it to the `StorageCluster StorageKind = "storagecluster.ocs.openshift.io/v1"`

Moreover, this is a variable for `kind` in our `spec`. For which we do need the same type to define all the vendors. https://github.com/red-hat-data-services/odf-operator/blame/main/api/v1alpha1/storagesystem_types.go#L60

jarrpa (approved):
/lgtm

PR #20, opened 2021-06-30
action: Add commitlint action for checking commits

iamniting:
commitlint will force a developer to write a commit as per our guidelines

reference links:
Naming Conventions -> https://commitlint.js.org/#/concepts-commit-conventions                                                                                                   
Rules -> https://commitlint.js.org/#/reference-rules

### DIFF
--- a/.github/workflows/commitlint.yaml
+++ b/.github/workflows/commitlint.yaml
@@ -1,0 +1,27 @@
+name: commitlint
+
+on:
+  push:
+    branches: [ '*' ]
+  pull_request:
+    branches: [ '*' ]
+
+jobs:
+  commitlint:
+    name: commitlint
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: run commitlint
+      uses: wagoid/commitlint-github-action@v3
+      with:
+        configFile: './.github/workflows/conf/commitlintrc.json'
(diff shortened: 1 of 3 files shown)


iamniting:
@jarrpa Let's merge this first so we can run this action on other PR's as well

jarrpa (changes requested):
I thiiiink it looks fine? Unfortunately I don't any information to what commitlint is, where it's from, how it's configured, or what the parameters we're configuring specify. Also this seems like it's going to create a `conf` directory when we already have a `config` directory? That doesn't make much sense. Can the JSON config not live with the workflow YAML?

iamniting:
> I thiiiink it looks fine? Unfortunately I don't any information to what commitlint is, where it's from, how it's configured, or what the parameters we're configuring specify. Also this seems like it's going to create a `conf` directory when we already have a `config` directory? That doesn't make much sense. Can the JSON config not live with the workflow YAML?

commitlint makes sure that the commit message is as per our guidelines, here is the reference https://commitlint.js.org/#/reference-rules.

I have created a conf directory inside the `.github/workflows/` to place all action related conf files

iamniting:
Here are some examples
Screenshot from 2021-07-02 14-23-14
Screenshot from 2021-07-02 14-20-39
Screenshot from 2021-07-02 14-25-33

iamniting:
@jarrpa I am merging this one myself. We can change this later if required.

PR #24, opened 2021-07-08
stop running webhook as a manager

iamniting:
To run a webhook via OLM it has a technical requirement that webhooks can not have the API groups same as OLM. So technically we can not contain the subscription webhook and its requirements in the CSV. remove the same from the CSV.

Error seen:
Webhook rules cannot include the OLM group

workarounds i can think of:
We might be able to create a job for this.
we can create a webhook deployment in the controller itself.

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -158,10 +158,6 @@ spec:
                   initialDelaySeconds: 15
                   periodSeconds: 20
                 name: manager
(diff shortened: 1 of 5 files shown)


iamniting:
It runs perfectly after removing the webhook from the manager

```
$ oc get pods
NAME                                               READY   STATUS    RESTARTS   AGE
odf-operator-controller-manager-5dd7b4bb75-rrglw   2/2     Running   0          4h14m
```

jarrpa (approved):
LGTM

PR #30, opened 2021-07-19
Add metrics for StorageSystem spec

umangachapagain:
Add metrics for StorageSystem spec so that we can use Prometheus queries to understand object relationships.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ COPY controllers/ controllers/
 COPY pkg/ pkg/
 COPY webhook/ webhook/
 COPY config/ config/
+COPY metrics/ metrics/
 
 # Run tests and linting
 RUN make test
--- a/controllers/storagesystem_controller.go
+++ b/controllers/storagesystem_controller.go
@@ -33,6 +33,7 @@ import (
 
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	odfv1alpha1 "github.com/red-hat-data-services/odf-operator/api/v1alpha1"
+	"github.com/red-hat-data-services/odf-operator/metrics"
 	"github.com/red-hat-data-services/odf-operator/pkg/util"
 )
 
@@ -75,6 +76,8 @@ func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	logger.Info("storagesystem instance found")
+
+	metrics.ReportODFSystemMapMetrics(instance.Name, instance.Spec.Name, instance.Spec.Namespace, string(instance.Spec.Kind))
 
 	if err := r.validateStorageSystemSpec(instance, logger); err != nil {
 		logger.Error(err, "failed to validate storagesystem")
--- a/metrics/metrics.go
+++ b/metrics/metrics.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2021 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	storageSystemMap = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "odf",
+		Subsystem: "",
+		Name:      "system_map",
+		Help:      "Map of ODF StorageSystems to their target Custom Resource",
+	}, []string{"storage_system", "target_name", "target_namespace", "target_kind"})
+)
+
+func init() {
+	metrics.Registry.MustRegister(storageSystemMap)
+}
+
+func ReportODFSystemMapMetrics(storageSystem, name, namespace, kind string) {
+	storageSystemMap.With(prometheus.Labels{
+		"storage_system":   storageSystem,
+		"target_name":      name,
+		"target_namespace": namespace,
+		"target_kind":      kind,
+	}).Set(1)
+}
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2021 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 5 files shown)


iamniting:
@umangachapagain you need to copy metrics into Dcokerfile. That's the reason for failure.

metrics/metrics_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var defaultRegistry = metrics.Registry
```
jarrpa, line 12: I see that you call `RegisterMetrics()` twice in this file, both of which act against `metrics.Registry` (as far as I understand). Is this not a problem, and if so why not?

metrics/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+	return prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "odf",
+		Name:      systemMapKey,
+		Help:      "Map of ODF Storage System to Subsystems",
+	}, []string{"system", "subsystem", "kind"})
+}
+
+func ReportODFSystemMapMetrics(system, subsystem, kind string) {
+	metrics, _ := storageSystemMetrics[systemMapKey].(*prometheus.GaugeVec)
+	metrics.WithLabelValues(system, subsystem, kind).Set(1)
```
jarrpa, line 39: I don't entirely understand what this function is doing. I'm assuming this is what records the actual mapping values, but what does `Set(1)` mean? I don't need this covered in the PR or comments, just a quick explanation here.

metrics/metrics.go
```diff
@@ -0,0 +1,40 @@
...
+	storageSystemMetrics := make(map[string]prometheus.Collector, 1)
+	storageSystemMetrics[systemMapKey] = newODFSystemMapMetrics()
+	return storageSystemMetrics
+}
+
+func newODFSystemMapMetrics() *prometheus.GaugeVec {
+	return prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "odf",
+		Name:      systemMapKey,
+		Help:      "Map of ODF Storage System to Subsystems",
```
jarrpa, line 33:
```suggestion
		Help:      "Map of ODF StorageSystems to their target CRs",
```

Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY api/ api/
 COPY controllers/ controllers/
 COPY pkg/ pkg/
 COPY webhook/ webhook/
-
```
jarrpa, line 18: Don't remove this newline, keep it beneath your new COPY statement.

jarrpa (changes requested):
Overall the logic looks solid. Left a few comments and questions. Speaking broadly, we don't have a formal notion of "subsystems" in the API so I don't want to introduce that terminology just for the metrics. Change instances of "subsystem" to "target" or something similar. A StorageCluster is not "part" of its StorageSystem, it *is* the StorageSystem.

metrics/metrics.go
umangachapagain: It's setting the actual metric value (which can't be a map). The mapping itself is stored as labels of the metric.
And it's set to 1 so that Prometheus queries become simpler.

metrics/metrics_test.go
umangachapagain: Removed the calls in favor of cleaner implementation.

It wasn't a problem because `RegisterMetrics()` was acting against `metrics.Registry` only once. Consequent calls would just return realizing that it was already called before.

metrics/metrics_test.go
```diff
@@ -0,0 +1,52 @@
...
+var find string = "odf_system_map"
+
+func TestReportODFSystemMapMetrics(t *testing.T) {
+	type args struct {
+		storageSystem, name, namespace, kind string
+	}
+	tests := []struct {
+		name string
```
iamniting, line 19: can you pls make this `name` as a `label`. just want to follow the same pattern everywhere.

metrics/metrics_test.go
```diff
@@ -0,0 +1,52 @@
...
+			name: "StorageSystem to Flashsystem",
+			args: args{
+				storageSystem: "StorageSystem2",
+				name:          "FlashSystem1",
+				namespace:     "Namespace1",
+				kind:          "Flashsystem",
+			},
+		},
+	}
+	for n, tt := range tests {
```
iamniting, line 41: Can you print the label here something like `t.Logf("Case %d: %s\n", i+1, tc.label)`

controllers/storagesystem_controller.go
```diff
@@ -80,6 +81,8 @@ func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	logger.Info("storagesystem instance found")
 
+	metrics.ReportODFSystemMapMetrics(instance.Name, instance.Spec.Name, instance.Spec.NameSpace, string(instance.Spec.Kind))
```
iamniting, line 84: Do we need this every reconcile? Can not we move this to if condition where we are adding a finalizer. So we can just call it for the first time.

iamniting (changes requested):

metrics/metrics_test.go
umangachapagain: t.Run() prints that for us.

controllers/storagesystem_controller.go
umangachapagain: If `Spec` is immutable, then we could do that. Else, we need this on every reconcile to account for changes.

obnoxxx (changes requested):
I have to say that I was initially very confused by the description and commit message (*transform* storagesystem spec into metrics) - I thought that you want to get rid of the storagesystem spec and replace it by metrics. Instead you are establishing metrics for the storage system. I would use "translate" instead of "transform" or say "Add metrics for the storagesystem spec" or similar.

metrics/metrics_test.go
jarrpa: Correct. Here's a good rundown: https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks

The output looks like this:
```
=== RUN   TestReportODFSystemMapMetrics
=== RUN   TestReportODFSystemMapMetrics/StorageSystem_to_StorageCluster
=== RUN   TestReportODFSystemMapMetrics/StorageSystem_to_Flashsystem
--- PASS: TestReportODFSystemMapMetrics (0.00s)
    --- PASS: TestReportODFSystemMapMetrics/StorageSystem_to_StorageCluster (0.00s)
    --- PASS: TestReportODFSystemMapMetrics/StorageSystem_to_Flashsystem (0.00s)
PASS
coverage: 100.0% of statements
ok  	github.com/red-hat-data-services/odf-operator/metrics	0.009s	coverage: 100.0% of statements
```

jarrpa (approved):
/lgtm

metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
+package metrics
```
iamniting, line 1: Can you pls add https://github.com/red-hat-storage/odf-operator/blob/main/hack/boilerplate.go.txt

metrics/metrics_test.go
```diff
@@ -0,0 +1,52 @@
+package metrics
```
iamniting, line 1: here as well https://github.com/red-hat-storage/odf-operator/blob/main/hack/boilerplate.go.txt

iamniting (changes requested):

iamniting (approved):

PR #36, opened 2021-07-26
Add initialization code for ODF Console

bipuladh:
- Creates a Deployment which contains the Web Server
- Creates a Service to expose the above Deployment
- Creates a ConsolePlugin to expose ODF Console Plugin to the OCP Core console. 

cc @jarrpa @umangachapagain @iamniting

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -170,6 +170,55 @@ spec:
           - create
         serviceAccountName: odf-operator-controller-manager
       deployments:
+      - name: odf-console
+        spec:
+          selector:
+            matchLabels:
+              app: odf-console
+          strategy: {}
+          template:
+            metadata:
+              labels:
+                app: odf-console
+            spec:
+              containers:
+              - args:
+                - --ssl --cert=/var/serving-cert/tls.crt --key=/var/serving-cert/tls.key
+                image: badhikar/odf-console:latest
+                name: odf-console
+                ports:
+                - containerPort: 9001
+                  protocol: TCP
+                resources:
+                  limits:
+                    cpu: 100m
+                    memory: 512Mi
+                volumeMounts:
+                - mountPath: /var/serving-cert
+                  name: odf-console-serving-cert
+                  readOnly: true
+              - args:
+                - --ssl --cert=/var/serving-cert/tls.crt --key=/var/serving-cert/tls.key
+                image: ibmcom/ibm-storage-odf-plugin:0.2.0
+                name: ibm-console
+                ports:
+                - containerPort: 9002
+                  protocol: TCP
+                resources:
+                  limits:
+                    cpu: 100m
+                    memory: 512Mi
+                volumeMounts:
+                - mountPath: /var/serving-cert
+                  name: ibm-console-serving-cert
+                  readOnly: true
+              volumes:
+              - name: odf-console-serving-cert
+                secret:
+                  secretName: odf-console-serving-cert
+              - name: ibm-console-serving-cert
+                secret:
+                  secretName: ibm-console-serving-cert
       - name: odf-operator-controller-manager
         spec:
           replicas: 1
@@ -198,6 +247,8 @@ spec:
                 - --health-probe-bind-address=:8081
                 - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
+                - --odf-console-port=9001
+                - --odf-console-port=9002
                 command:
                 - /manager
                 env:
--- a/console/console.go
+++ b/console/console.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2021 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package console
+
+import (
+	"context"
+
+	consolev1alpha1 "github.com/openshift/api/console/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const DEPLOYMENT_NAMESPACE = "openshift-storage"
+
+func GetService(serviceName string, port int, owner metav1.ObjectMeta) apiv1.Service {
+	return apiv1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceName + "-service",
+			Namespace: DEPLOYMENT_NAMESPACE,
+			Annotations: map[string]string{
+				"service.alpha.openshift.io/serving-cert-secret-name": serviceName + "-serving-cert",
+			},
+			Labels: map[string]string{
+				"app": "odf-ui",
+			},
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: "apps/v1",
+					Kind:       "Deployment",
+					Name:       "odf-console",
+					UID:        owner.UID,
+				},
+			},
+		},
+		Spec: apiv1.ServiceSpec{
+			Ports: []apiv1.ServicePort{
+				{Protocol: "TCP",
+					TargetPort: intstr.IntOrString{IntVal: int32(port)},
+					Port:       int32(port),
+					Name:       "console-port",
+				},
+			},
+			Selector: map[string]string{
+				"app": "odf-ui",
+			},
+			Type: "ClusterIP",
+		},
+	}
+}
+
+func GetConsolePluginCR(displayName string, consolePort int, serviceName string, owner metav1.ObjectMeta) consolev1alpha1.ConsolePlugin {
+	return consolev1alpha1.ConsolePlugin{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "odf-console",
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: "apps/v1",
+					Kind:       "Deployment",
+					Name:       "odf-console",
+					UID:        owner.UID,
+				},
+			},
+		},
+		Spec: consolev1alpha1.ConsolePluginSpec{
+			DisplayName: displayName,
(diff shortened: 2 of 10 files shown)


iamniting:
@bipuladh Can you pls take a look at failures?

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+			Service: consolev1alpha1.ConsolePluginService{
+				Name:      "odf-service",
+				Namespace: "openshift-storage",
+				Port:      int32(9001),
+			},
+		},
+	}
+	service = &apiv1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "odf-service",
```
iamniting, line 86:
```suggestion
			Name:      "odf-console-service",
```

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	consolePlugin = &consolev1alpha1.ConsolePlugin{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "odf-console",
+		},
+		Spec: consolev1alpha1.ConsolePluginSpec{
+			DisplayName: "ODF Console",
+			Service: consolev1alpha1.ConsolePluginService{
+				Name:      "odf-service",
+				Namespace: "openshift-storage",
```
iamniting, line 79:
```suggestion
				Name:      "odf-console-service",
				Namespace: DEPLOYMENT_NAMESPACE,
```

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+			Ports: []apiv1.ServicePort{
+				{Name: "jack", Port: int32(9001)},
+			},
+		},
+	}
+)
+
+func ConsoleInit(client client.Client) error {
+	// Create a deployment for the UI WebServer
+	err := client.Create(context.TODO(), serverDeployment)
```
iamniting, line 99: When this code runs for the first time it will pass, but it will never pass next time onwards as you are not handling the case if the resources are already present on the cluster.

console/main.go
```diff
@@ -0,0 +1,114 @@
+package console
```
iamniting, line 1: can we rename this file to something else instead of main.go. as there is no main func in this file

iamniting (changes requested):

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		Spec: apiv1.ServiceSpec{
+			Ports: []apiv1.ServicePort{
+				{Name: "jack", Port: int32(9001)},
+			},
+		},
+	}
+)
+
+func ConsoleInit(client client.Client) error {
```
umangachapagain, line 97:
```suggestion
func Initialize(client client.Client) error {
```
Since package is imported as console, this will translate to `console.Initialize` which is better.

main.go
```diff
@@ -37,6 +37,7 @@ import (
 
 	//+kubebuilder:scaffold:imports
 	consolev1 "github.com/openshift/api/console/v1"
+	console "github.com/red-hat-storage/odf-operator"
```
umangachapagain, line 40:
```suggestion
	console "github.com/red-hat-data-services/odf-operator/console"
```

umangachapagain (changes requested):

umangachapagain (commented):
This approach is deploy and forget.
If you need to ensure that web server always exists, you need to add some more code and configuration.

console/main.go
jarrpa: +1

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const DEPLOYMENT_NAME = "odf-console"
+const DEPLOYMENT_NAMESPACE = "openshift-storage"
+const ODF_CONSOLE_IMAGE = "badhikar/odf-console:latest"
```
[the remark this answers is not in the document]
jarrpa: `deploymentName` can be hardcoded, I'm not too concerned about that one. The rest yes, they should be configurable values. Just do something like this: https://github.com/openshift/ocs-operator/blob/8faa2bf23e7c5b87a242206dfd26df32516c55e5/controllers/storagecluster/storagecluster_controller.go#L30-L32
For the Namespace we already have a variable available: https://github.com/openshift/ocs-operator/blob/master/controllers/util/k8sutil.go

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+var (
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(consolev1.AddToScheme(scheme))
+	utilruntime.Must(consolev1alpha1.Install(scheme))
+	//+kubebuilder:scaffold:scheme
+}
```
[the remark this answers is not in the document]
jarrpa: +1

console/main.go
```diff
@@ -0,0 +1,114 @@
...
+					Containers: []apiv1.Container{
+						{
+							Name:  "odf-console",
+							Image: ODF_CONSOLE_IMAGE,
+							Ports: []apiv1.ContainerPort{
+								{
+									Name:          "http",
+									Protocol:      apiv1.ProtocolTCP,
+									ContainerPort: 9001,
+								},
```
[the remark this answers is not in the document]
jarrpa: Yes, this one should probably be a CLI option rather than an env var.

console/main.go
jarrpa: +1

main.go
jarrpa: Don't change to `data-services`, we've just moved away from that. But yet to the `console` addition.

jarrpa (changes requested):

umangachapagain:
> Don't change to data-services, we've just moved away from that. But yet to the console addition.
>
@jarrpa yes, but the module still refers to "data-services". So Go will complain.

https://github.com/red-hat-storage/odf-operator/blob/b3f0656442c86cf126f3a089e84d388737afc64f/go.mod#L1

console/main.go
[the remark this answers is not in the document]
bipuladh: > > Don't change to data-services, we've just moved away from that. But yet to the console addition.
> 
> @jarrpa yes, but the module still refers to "data-services". So Go will complain.
> 
> https://github.com/red-hat-storage/odf-operator/blob/b3f0656442c86cf126f3a089e84d388737afc64f/go.mod#L1

@umangachapagain it complained. Bad go!

jarrpa:
> > Don't change to data-services, we've just moved away from that. But yet to the console addition.
> 
> @jarrpa yes, but the module still refers to "data-services". So Go will complain.
> 
> https://github.com/red-hat-storage/odf-operator/blob/b3f0656442c86cf126f3a089e84d388737afc64f/go.mod#L1

We should probably update that, then.

main.go
```diff
@@ -39,6 +39,7 @@ import (
 
 	//+kubebuilder:scaffold:imports
 	consolev1 "github.com/openshift/api/console/v1"
+	consolev1alpha1 "github.com/openshift/api/console/v1alpha1"
```
iamniting, line 42: do we really need 2 versions of the console can not it be done with one.

main.go
```diff
@@ -122,4 +128,14 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+
+	setupLog.Info("starting console")
```
iamniting, line 132: I think we should call this controller above where we are calling other controllers in the same file.

controllers/console_controller.go
```diff
@@ -0,0 +1,135 @@
+package controllers
```
iamniting, line 1: Question: Did you created a this file via operator-sdk scafolding?
This file should also have https://github.com/red-hat-storage/odf-operator/blob/main/hack/boilerplate.go.txt

controllers/console_controller.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"context"
+	"strings"
+
+	"github.com/go-logr/logr"
+	consolev1alpha1 "github.com/openshift/api/console/v1alpha1"
```
iamniting, line 8: can we separate imports like https://github.com/red-hat-storage/odf-operator/blob/main/controllers/subscriptions.go#L19

console/console.go
```diff
@@ -0,0 +1,103 @@
+package console
+
+import (
+	consolev1alpha1 "github.com/openshift/api/console/v1alpha1"
```
iamniting, line 4: can we separate imports like https://github.com/red-hat-storage/odf-operator/blob/main/controllers/subscriptions.go#L19

bundle/manifests/odf-operator.clusterserviceversion.yaml
```diff
@@ -198,6 +198,8 @@ spec:
                 - --health-probe-bind-address=:8081
                 - --metrics-bind-address=127.0.0.1:8080
                 - --leader-elect
+                - --console-image=badhikar/odf-console:latest
```
iamniting, line 201: This should also go in the config/manager/manager.yaml But I think we should have this as an ENV variable.

console/console.go
```diff
@@ -0,0 +1,103 @@
+package console
```
iamniting, line 1: This file should also have https://github.com/red-hat-storage/odf-operator/blob/main/hack/boilerplate.go.txt

iamniting (changes requested):

controllers/console_controller.go
```diff
@@ -0,0 +1,135 @@
...
+func (r *ConsoleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Log.WithValues("odf-console", req.NamespacedName)
+	// Create deployment if doesn't exist
+	deployment := &appsv1.Deployment{}
+	idealDeployment := console.GetDeployment(r.ConsoleImage, r.ConsolePort)
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: idealDeployment.ObjectMeta.Name, Namespace: idealDeployment.ObjectMeta.Namespace}, deployment); errors.IsNotFound(err) {
+		if err := r.Client.Create(ctx, &idealDeployment); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
aruniiird, line 48: Please check the below comment for `sevrice` variable which applicable for `deployment` var.

aruniiird (changes requested):
Some minor comments

iamniting:
@bipuladh Can you separate this one commit into multiple commits and do something like #56 to feed images via makefile.

console/console.go
bipuladh: ack done.

main.go
bipuladh: Yes at least for now. As dynamic plugins API are still in alpha and have not made it as part of v1 stable.

jarrpa: Good to know it's at least not our fault, I guess? :joy:

jarrpa (approved):
Well... I hate to say but I have some problems with this. That said I'll wave it through for now just so we can get it into the builds. ;) Please submit a followup PR for the following:

1. Define `odf-console` as a global variable and reference it as needed
2. Please add some unit tests!! :stuck_out_tongue: 

/lgtm

PR #38, opened 2021-07-27
Add bundle dependency

iamniting:
Add "ocs-operator" as a dependency of odf

Add "ocs-operator" and "ibm-storage-odf-operator" bundle images to build
within our catalog image.

### DIFF
--- a/bundle/metadata/dependencies.yaml
+++ b/bundle/metadata/dependencies.yaml
@@ -1,0 +1,5 @@
+dependencies:
+- type: olm.package
+  value:
+    packageName: ocs-operator
+    version: ">=4.9.0 <4.10.0"
--- a/hack/make-bundle-vars.mk
+++ b/hack/make-bundle-vars.mk
@@ -44,9 +44,12 @@ CATALOG_IMG ?= $(IMAGE_REGISTRY)/$(REGISTRY_NAMESPACE)/$(CATALOG_IMAGE_NAME):$(I
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 
+
 # A comma-separated list of bundle images (e.g. make catalog-build BUNDLE_IMGS=example.com/operator-bundle:v0.1.0,example.com/operator-bundle:v0.2.0).
 # These images MUST exist in a registry and be pull-able.
-BUNDLE_IMGS ?= $(BUNDLE_IMG)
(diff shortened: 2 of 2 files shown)


bundle/metadata/dependencies.yaml
```diff
@@ -0,0 +1,5 @@
+dependencies:
+- type: olm.package
+  value:
+    packageName: ocs-operator
+    version: ">=4.7.0 <=4.9.0"
```
b-ranto, line 5: I am little worried about this line. Will this be able to pull a version like `4.9.0-2.ci`?

iamniting: YOu can read more about versions here https://olm.operatorframework.io/docs/concepts/olm-architecture/dependency-resolution/#declaring-dependencies

b-ranto: Hmm, I am still wondering. Could it make more sense to have something like `>=4.9.0 < 4.10.0` there? This could try to use ocs-operator from 4.8 or 4.7. Is that what we really want?

iamniting: Done

hack/make-bundle-vars.mk
```diff
@@ -46,7 +46,9 @@ CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
 
 # A comma-separated list of bundle images (e.g. make catalog-build BUNDLE_IMGS=example.com/operator-bundle:v0.1.0,example.com/operator-bundle:v0.2.0).
 # These images MUST exist in a registry and be pull-able.
-BUNDLE_IMGS ?= $(BUNDLE_IMG)
+OCS_BUNDLE_IMG ?= quay.io/ocs-dev/ocs-operator-bundle:latest
+IBM_BUNDLE_IMG ?= docker.io/ibmcom/ibm-storage-odf-operator-bundle:0.2.0
+BUNDLE_IMGS ?= $(BUNDLE_IMG),$(OCS_BUNDLE_IMG),$(IBM_BUNDLE_IMG)
```
jarrpa, line 51: Can we do something to optionally concatenate these strings? It's possible we may want to sometimes generate the catalog index with only odf-operator bundles.

iamniting: What should be the default behavior?

jarrpa: We should also be more careful on which tags of the bundle images we want (e.g. something other than `latest` for ocs-operator without having to write the entire URL), but that can also be a later PR.

iamniting: 1, 2, 3 are done

jarrpa (approved):
/lgtm

PR #47, opened 2021-08-03
controllers: use env var instead of hardcoded values

iamniting:
remove the constant values and define a map of default values. Which
will be used only if env var is not set/defined.


jarrpa (approved):
/lgtm

PR #48, opened 2021-08-03
make some overall enhancements 

iamniting:
Signed-off-by: Nitin Goyal nigoyal@redhat.com


config/manager/manager.yaml
```diff
@@ -45,12 +45,5 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
-        resources:
-          limits:
-            cpu: 100m
-            memory: 30Mi
-          requests:
-            cpu: 100m
-            memory: 20Mi
```
jarrpa, line 54: Why are you doing this? Apart from affecting our QoS class, certain environments (hi Dedicated!) require all our Pods to define explicit resource consumption.

jarrpa (changes requested):
For the first two commits: don't split api and controller changes. In general we should try to make sure each commit leaves the code in a valid state, and splitting these would prevent the code from building. Also typo in the commit message: `s/feild/field/`

The subsequent bundle commits can also be merged, just mention both commands you ran.

config/manager/manager.yaml
iamniting: Actually, the pod was getting oomKilled because of which I just removed them. But now I have increased the limit.

iamniting:
> For the first two commits: don't split api and controller changes. In general we should try to make sure each commit leaves the code in a valid state, and splitting these would prevent the code from building. Also typo in the commit message: `s/feild/field/`
> 
> The subsequent bundle commits can also be merged, just mention both commands you ran.

Done

jarrpa (approved):
/lgtm

PR #49, opened 2021-08-04
action: Add action to wait for dependent PR to be merged

iamniting:
We can block the PR from merging and wait for its dependencies to be merged via adding `Depends on #<>`


iamniting:
Depends on #48

PR #52, opened 2021-08-09
controllers: Fix scheme issue via adding ibm scheme

iamniting:
Fix: no kind is registered for the type v1alpha1.FlashSystemCluster
in scheme.


iamniting:
Approved it myself as this is a very small fix and does not really require a review.

PR #67, opened 2021-08-20
bundle: Add annotation to indentify ODF contains a dynamic plugin.

bipuladh:
- Enables a checkbox in the UI that would allow users to enable the UI plugin.

Since ODF is part of `redhat-operators` catalog this checkbox will be selected by default. Not adding IBM as we don't want to enable IBM UI by default. 

/assign @iamniting @jarrpa


bipuladh:
/close 
Superseded by https://github.com/red-hat-storage/odf-operator/pull/68

PR #68, opened 2021-08-20
bundle: Update CSV to use official console image and add console annotation

bipuladh:
- Use official images for ODF Console from quay repository. 
- Update CSV annotations to identify ODF Operator as a dynamic plugin carrier.


iamniting (approved):
/lgtm

PR #71, opened 2021-08-20
controllers: Add and update conditions in storageSystem controller

iamniting:
Mark storage system as ready if csv is ready, either underlying storage
CR is present or not, so that UI can create a IBM flashsystem based on
its phase.

Without this change UI will never be able to create a IBM flashsystem.

### DIFF
--- a/api/v1alpha1/storagesystem_types.go
+++ b/api/v1alpha1/storagesystem_types.go
@@ -30,8 +30,14 @@ import (
 type StorageKind string
 
 const (
-	// ConditionResourcePresent communicates the status of underlying resource
-	ConditionResourcePresent conditionsv1.ConditionType = "ResourcePresent"
+	//ConditionStorageSystemInvalid communicates if storagesystem CR is invalid
+	ConditionStorageSystemInvalid conditionsv1.ConditionType = "StorageSystemInvalid"
+
+	// ConditionVendorCsvReady communicates if CSV is ready
+	ConditionVendorCsvReady conditionsv1.ConditionType = "VendorCsvReady"
+
+	// ConditionVendorSystemPresent communicates if backend storage CR is present in the cluster
+	ConditionVendorSystemPresent conditionsv1.ConditionType = "VendorSystemPresent"
 )
 
 // StorageSystemSpec defines the desired state of StorageSystem
--- a/api/v1alpha1/storagesystem_types.go
+++ b/api/v1alpha1/storagesystem_types.go
@@ -56,7 +56,7 @@ type StorageSystemSpec struct {
 type StorageSystemStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// Conditions describes the state of the StorageCluster resource.
+	// Conditions describes the state of the StorageSystem resource.
 	// +optional
 	Conditions []conditionsv1.Condition `json:"conditions,omitempty"`
 }
--- a/api/v1alpha1/storagesystem_types.go
+++ b/api/v1alpha1/storagesystem_types.go
@@ -34,20 +34,6 @@ const (
 	ConditionResourcePresent conditionsv1.ConditionType = "ResourcePresent"
 )
 
-const (
-	// PhaseDeleting represents the Deleting state of storagesystem
-	PhaseDeleting = "Deleting"
-
-	// PhaseError represents the Error state of storagesystem
-	PhaseError = "Error"
-
-	// PhaseProgressing represents the Progressing state of storagesystem
-	PhaseProgressing = "Progressing"
-
-	// PhaseReady represents the Ready state of storagesystem
-	PhaseReady = "Ready"
-)
-
 // StorageSystemSpec defines the desired state of StorageSystem
 type StorageSystemSpec struct {
 	// Important: Run "make" to regenerate code after modifying this file
@@ -70,11 +56,6 @@ type StorageSystemSpec struct {
 type StorageSystemStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// Phase describes the Phase of StorageSystem
-	// This is used by OLM UI to provide status information
-	// to the user
-	Phase string `json:"phase,omitempty"`
-
 	// Conditions describes the state of the StorageCluster resource.
 	// +optional
 	Conditions []conditionsv1.Condition `json:"conditions,omitempty"`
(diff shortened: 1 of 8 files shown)


controllers/conditions.go
```diff
@@ -17,60 +17,135 @@ limitations under the License.
...
+		Type:    conditionsv1.ConditionProgressing,
+		Status:  corev1.ConditionTrue,
+		Reason:  reason,
+		Message: message,
+	})
 }
 
-func GetConditionResourcePresent(status corev1.ConditionStatus, reason, msg string) conditionsv1.Condition {
-	return GetCondition(odfv1alpha1.ConditionResourcePresent, status, reason, msg)
+func SetReconcileAccomplishCondition(conditions *[]conditionsv1.Condition,
```
jarrpa, line 79:
```suggestion
func SetReconcileCompleteConditions(conditions *[]conditionsv1.Condition,
```

controllers/conditions.go
```diff
@@ -17,60 +17,135 @@ limitations under the License.
...
+	})
+	conditionsv1.SetStatusCondition(conditions, conditionsv1.Condition{
+		Type:    conditionsv1.ConditionProgressing,
+		Status:  corev1.ConditionTrue,
+		Reason:  reason,
+		Message: message,
+	})
+}
+
+func SetStorageSystemInValidCondition(conditions *[]conditionsv1.Condition,
```
jarrpa, line 113: Treat this function like you do the other two. Only focus on one condition and take the status as a parameter. In particular, you should be setting invalid=false when we have validated the spec.

controllers/conditions.go
```diff
@@ -17,60 +17,135 @@ limitations under the License.
...
+	reason string, message string) {
+
+	conditionsv1.SetStatusCondition(conditions, conditionsv1.Condition{
+		Type:    conditionsv1.ConditionProgressing,
```
jarrpa, line 117: We can still consider this to be true. If the spec is invalid then reconciliation is not complete. It would be more appropriate to say ConditionUnknown, since we can't even check. The only time we should be setting this to false is when reconciliation completes successfully.

controllers/conditions.go
```diff
@@ -17,60 +17,135 @@ limitations under the License.
...
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
-	ocsv1 "github.com/openshift/ocs-operator/api/v1"
 	odfv1alpha1 "github.com/red-hat-data-services/odf-operator/api/v1alpha1"
 )
 
-func GetCondition(condition conditionsv1.ConditionType, status corev1.ConditionStatus, reason, msg string) conditionsv1.Condition {
-	return conditionsv1.Condition{
-		Type:    condition,
-		Status:  status,
+func SetReconcileInitCondition(conditions *[]conditionsv1.Condition,
```
jarrpa, line 26:
```suggestion
func SetReconcileInitConditions(conditions *[]conditionsv1.Condition,
```

api/v1alpha1/storagesystem_types.go
```diff
@@ -30,8 +30,14 @@ import (
 type StorageKind string
 
 const (
-	// ConditionResourcePresent communicates the status of underlying resource
-	ConditionResourcePresent conditionsv1.ConditionType = "ResourcePresent"
+	//ConditionStorageSystemInValid communicates if storagesystem CR is inValid
+	ConditionStorageSystemInValid conditionsv1.ConditionType = "StorageSystemInValid"
```
jarrpa, line 34: Here and in every other instance: the v in invalid should never be capitalized, it's one word and not two. Look for both `InValid` and `inValid`.

jarrpa (changes requested):

jarrpa:
Overall much better. Our discussions were not in vain! The requested changes are mostly typographical, but there's a few for slightly changing the scope and use of the set condition functions.

controllers/conditions.go
iamniting: Done

api/v1alpha1/storagesystem_types.go
iamniting: Done

controllers/conditions.go
iamniting: Done

controllers/conditions.go
```diff
@@ -17,60 +17,135 @@ limitations under the License.
...
-		conditionsv1.SetStatusCondition(&instance.Status.Conditions, GetConditionResourcePresent(corev1.ConditionFalse, "NotFound", err.Error()))
-		requeue = true
-	} else {
-		logger.Error(err, "set condition", string(odfv1alpha1.ConditionResourcePresent), corev1.ConditionUnknown)
-		conditionsv1.SetStatusCondition(&instance.Status.Conditions, GetConditionResourcePresent(corev1.ConditionUnknown, "", err.Error()))
-		requeue = true
-	}
-
-	return requeue, err
+func SetDeleteionInProgressCondition(conditions *[]conditionsv1.Condition,
```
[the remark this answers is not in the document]
iamniting: Done

controllers/conditions.go
iamniting: Done

controllers/conditions.go
iamniting: In that case, According to that, we do need to change progress also.

iamniting:
> Overall much better. Our discussions were not in vain! The requested changes are mostly typographical, but there's a few for slightly changing the scope and use of the set condition functions.

@jarrpa I have added one more commit at last to simplify conditions other than what you requested. Pls do take a look. If you like it let me know otherwise I will drop this.

controllers/conditions.go
```diff
@@ -23,129 +23,110 @@ import (
...
+		conditions, corev1.ConditionFalse, reason, message)
+	SetProgressingCondition(
+		conditions, corev1.ConditionTrue, reason, message)
+
+	SetStorageSystemInvalidCondition(
+		conditions, corev1.ConditionUnknown, reason, message)
+	SetVendorCsvReadyCondition(
+		conditions, corev1.ConditionUnknown, reason, message)
+	SetVendorSystemPresentCondition(
+		conditions, corev1.ConditionUnknown, reason, message)
```
jarrpa, line 94: omg I didn't think about this but this is so much better, thank you!! I would prefer to not have the wrapping, but that's an aesthetic choice. Leave as-is if you like it better.

jarrpa (approved):
/lgtm

PR #75, opened 2021-08-30
bundle: Change odf-catalogsource namespace

iamniting:
create odf-catalogsource in openshift-storage namespace instead of
openshift-marketplace while installing via `make deploy-with-olm`.


jarrpa (approved):
/lgtm

PR #82, opened 2021-09-07
bundle: Add initialization-resource annotation in the csv

iamniting:
Add a annotation with the default storage system to create a storage
system exactly after installing the operator.

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -37,6 +37,20 @@ metadata:
     containerImage: quay.io/ocs-dev/odf-operator:latest
     description: OpenShift Data Foundation provides a common control plane for storage
       solutions on OpenShift Container Platform.
+    operatorframework.io/initialization-resource: |-
+      {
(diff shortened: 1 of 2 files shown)


iamniting:
Tested this out and it worked

Screenshot from 2021-09-07 18-34-55
Screenshot from 2021-09-07 18-38-14

bipuladh:
For some reason you're not being redirected to the right URL. 
Can you share the URL of the page you are getting redirected to?

iamniting:
> For some reason you're not being redirected to the right URL.
> Can you share the URL of the page you are getting redirected to?

Actually, I have tested this on a 4.8 cluster

bipuladh:
> > For some reason you're not being redirected to the right URL.
> > Can you share the URL of the page you are getting redirected to?
> 
> Actually, I have tested this on a 4.8 cluster

Ah okay!
/lgtm

iamniting:
@bipuladh It looks like there is a bug in the openshift-ci as you are also able to give /lgtm as your name is not in the owner's file. @jarrpa am I missing something here?

bipuladh:
Lgtm tags can be provided by anyone. I however cannot approve PRs

PR #86, opened 2021-09-13
Update IBM console and CSV version to 0.2.1

iamniting:


bipuladh (commented):
Maybe we should also move to quay at this point?

iamniting:
> Maybe we should also move to quay at this point?

That is on the IBM folks, we can bring this again in one of our meetings. But I don't want this PR to wait on it.

jarrpa (approved):
/lgtm

PR #89, opened 2021-09-13
makefile: Update locations of the dependent images

iamniting:
Update the location so that the values of these images wont get effect
while building the odf images. eg export REGISTRY_NAMESPACE=nigoyal
should not effect the odf-console or ocs-operator-bundle image anymore.

Moreover use seprate location variables for IBM bundle and IBM console
to provide a functaniolity to change even one of them only.


jarrpa (approved):
/lgtm

PR #90, opened 2021-09-14
bundle: fix ibm-console container args

shdn-ibm:
Signed-off-by: shdn@cn.ibm.com
As ibm-console container use `Entrypoint` in Dockerfile and does not include bash pkg to reduce image size, every argument of the container should be a separated line in tha yaml file.


iamniting:
@bipuladh Do we need to backport this if yes do we have a bug?

iamniting (changes requested):
Can you pls fix the DCO failure

also change the commit-msg to 
```
bundle: fix ibm-console container args
```

shdn-ibm:
The commit is broken after following the instruction in DCO 
Close this PR, and will raise another one...

PR #94, opened 2021-09-15
console: remove ibm-console container from odf-console deployement

bipuladh:
https://bugzilla.redhat.com/show_bug.cgi?id=2004824

/assign @jarrpa @iamniting


Makefile
```diff
@@ -102,7 +102,6 @@ bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metada
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	cd config/default && $(KUSTOMIZE) edit set image rbac-proxy=$(RBAC_PROXY_IMG)
 	cd config/console && $(KUSTOMIZE) edit set image odf-console=$(ODF_CONSOLE_IMG)
-	cd config/console && $(KUSTOMIZE) edit set image ibm-console=$(IBM_CONSOLE_IMG)
```
iamniting, line 105: Line no. 86 should be also removed.

config/console/console_init.yaml
```diff
@@ -22,30 +22,14 @@ spec:
             - containerPort: 9001
               protocol: TCP
           args:
-            - "--ssl --cert=/var/serving-cert/tls.crt --key=/var/serving-cert/tls.key"
+            - "--ssl"
+            - "--cert=/var/serving-cert/tls.crt"
+            - "--key=/var/serving-cert/tls.key"
```
iamniting, line 27: This should be a separate commit

main.go
```diff
@@ -73,14 +73,12 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	var odfConsolePort int
-	var ibmConsolePort int
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8085", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8082", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.IntVar(&odfConsolePort, "odf-console-port", 9001, "The port where the ODF console server will be serving it's payload")
```
iamniting, line 82: I would suggest you to keep the line as it was before and just delete the IBM related ones.

console/console.go
```diff
@@ -95,7 +95,7 @@ func GetConsolePluginCR(pluginName string, displayName string, consolePort int,
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=*
 
-func InitConsole(client client.Client, odfPort int, ibmPort int) error {
+func InitConsole(client client.Client, consolePort int) error {
```
iamniting, line 98: keep the name as it was before to minimize the changes.

iamniting (changes requested):
Screenshot from 2021-09-15 14-52-12
Also i would like to have something like this in the generated code.

config/console/console_init.yaml
bipuladh: It's a trivial change. Let's keep it as part of it.

console/console.go
bipuladh: The names don't make much sense. I think it's better to make variable names contextually correct.

iamniting: If thats a case lets do it at all the places even in main.go `odfConsolePort` should be `ConsolePort`

iamniting (changes requested):
Ok all changes look good to me except the commits. I would like you to correct commits.

iamniting (approved):
/lgtm

config/console/console_init.yaml
jarrpa: It is precisely because it is a trivial change that it should be moved out. :) Keep the commit tight and focused on only the relevant logical changes.

jarrpa (changes requested):
Overall this makes sense, however I kinda liked seeing the console port called "odf-console-port", since it was more explicit. Please revert those renaming changes.

config/console/console_init.yaml
bipuladh: So should I create a different back-port for this?

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -24,5 +24,4 @@ spec:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
-        - "--odf-console-port=9001"
-        - "--ibm-console-port=9003"
+        - "--console-port=9001"
```
obnoxxx, line 27: This patch removes ibm-console. If you want to rename the odf-console-port, then it should be separate commit because it is a logically separate change. It can be discussed

* whether this should remain as `--odf-console-port` for explicitness as @jarrpa  requested
* whether the renaming should even be a separate PR

But at the very least it needs to be a separate commit for starters.

console/console.go
obnoxxx: Even if the names don't make much sense, the renaming should be in a separate commit.

* one commit: remove ibm console
* one or more separate commits: do follow-up renames and cleanups

console/console.go
```diff
@@ -104,24 +104,14 @@ func InitConsole(client client.Client, odfPort int, ibmPort int) error {
 		return err
 	}
 	// Create core ODF console Service
-	odfService := GetService("odf-console", odfPort, deployment.ObjectMeta)
+	odfService := GetService("odf-console", consolePort, deployment.ObjectMeta)
```
obnoxxx: separate commit

console/console.go
```diff
@@ -104,24 +104,14 @@ func InitConsole(client client.Client, odfPort int, ibmPort int) error {
...
 	if err := client.Create(context.TODO(), &odfService); err != nil && !errors.IsAlreadyExists(err) {
 		return err
 	}
 	// Create core ODF Plugin
-	odfConsolePlugin := GetConsolePluginCR("odf-console", "ODF Plugin", odfPort, odfService.ObjectMeta.Name, deployment.ObjectMeta)
+	odfConsolePlugin := GetConsolePluginCR("odf-console", "ODF Plugin", consolePort, odfService.ObjectMeta.Name, deployment.ObjectMeta)
```
obnoxxx: separate commit

main.go
```diff
@@ -72,15 +72,13 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
-	var odfConsolePort int
-	var ibmConsolePort int
+	var consolePort int
```
obnoxxx: renaming separate commit (or leave as is)

main.go
obnoxxx: ack. one commit only removing, other commits can rename / cleanup etc.

main.go
```diff
@@ -104,7 +102,7 @@ func main() {
 
 	setupLog.Info("starting console")
 	if err := mgr.Add(manager.RunnableFunc(func(context.Context) error {
-		err = console.InitConsole(mgr.GetClient(), odfConsolePort, ibmConsolePort)
+		err = console.InitConsole(mgr.GetClient(), consolePort)
```
obnoxxx: rename a separate commit (or don't rename)

obnoxxx (changes requested):
It is really important to keep commits small and atomic.

* one commit to remove the ibm console
* one or more commits to rename variables / cmdline switches (but there is objection to it from @jarrpa )
* one or more commits to do unrelated cleanups like reformatting

See https://github.com/red-hat-storage/odf-operator/blob/main/CONTRIBUTING.md

obnoxxx (approved):
this is clean now, thanks! renamings and other cleanups can be done in a follow up PR.

/lgtm

iamniting:
@bipuladh Can you pls send this PR to the release-4.9 branch as well?

iamniting:
Backported PR is https://github.com/red-hat-storage/odf-operator/pull/98

PR #99, opened 2021-09-21
Manage dependency Subscriptions directly

jarrpa:
This PR removes the use of OLM bundle dependencies in favor of managing our subcomponent Subscriptions directly. This is to work around a limitation of the OLM where it will not update the channel of a Subscription to resolve a dependency conflict. We need this specifically because we use channels to separate our y-stream releases. In particular, this allows for a simple upgrade path from OCS 4.8 to ODF 4.9 by simply subscribing to ODF.

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

https://bugzilla.redhat.com/show_bug.cgi?id=2000143


controllers/subscriptions.go
```diff
@@ -153,3 +155,71 @@ func (r *StorageSystemReconciler) isVendorCsvReady(instance *odfv1alpha1.Storage
...
+			InstallPlanApproval:    operatorv1alpha1.ApprovalAutomatic,
+		},
+	}
+
+	return []*operatorv1alpha1.Subscription{noobaaSubscription, ocsSubscription}
+}
+
+// GetFlashSystemClusterSubscription return subscription for FlashSystemCluster
+func GetFlashSystemClusterSubscriptions() []*operatorv1alpha1.Subscription {
+	ibmSubscription := &operatorv1alpha1.Subscription{
```
iamniting, line 189: What about the IBM CSI subscription, Do we want to handle that also ourselves?

controllers/subscriptions.go
```diff
@@ -153,3 +135,71 @@ func (r *StorageSystemReconciler) isVendorCsvReady(instance *odfv1alpha1.Storage
 		return err
 	}
 }
+
+// GetSubscriptions returns all required Subscriptions for the given StorageKind
+func GetSubscriptions(k odfv1alpha1.StorageKind) []*operatorv1alpha1.Subscription {
```
iamniting, line 140: How about passing the instance itself to the func, It will give us the freedom to even change the logic afterwards without changing args.

controllers/storagesystem_controller.go
```diff
@@ -224,3 +225,35 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
 		Complete(r)
 }
+
+// RemoveSubscriptions deletes any managed Subscriptions that do not have any
+// existing CRs that their operators reconcile.
+func (r *StorageSystemReconciler) RemoveSubscriptions() error {
```
iamniting, line 231: Lets move this func to subscriptions.go

controllers/storagesystem_controller.go
```diff
@@ -224,3 +225,35 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				}
+			}
+		}
+	}
+
+	return err
+}
+
+// GetOwnerReferences returns an OwnerReference for the specified StorageSystem
+func GetOwnerReferences(instance *odfv1alpha1.StorageSystem) []metav1.OwnerReference {
```
iamniting, line 251: I would like to see this func before setupWithManager exactly after addReferenceToRelatedObjects. If possible let's covert this as a method on top of the reconciler.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.15 as builder
+FROM golang:1.16 as builder
```
iamniting, line 2: From now onwards let's use bundle as a type for commit msg for all DOCKEFILE related changes.

iamniting (changes requested):

Dockerfile
jarrpa: I can do that now

Makefile
```diff
@@ -47,8 +47,19 @@ test: manifests generate fmt vet ## Run tests.
 	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
 
 define MANAGER_ENV_VARS
-OCS_CSV_NAME=$(OCS_CSV_NAME)
+OPERATOR_NAMESPACE=openshift-storage
```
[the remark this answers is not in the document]
jarrpa: Oh, duh, forgot I was going to do this. :joy: Will do.

controllers/storagesystem_controller.go
jarrpa: ...oh duh, absolutely!

controllers/storagesystem_controller.go
jarrpa: There's no need to do that. In fact, I was rather thinking of getting rid of this entirely since I found a function that does exactly this: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference

controllers/subscriptions.go
jarrpa: I try to follow a principle of passing as little information as is strictly required. Changing a function signature is a trivial change with most IDEs, and in particular it will highlight all references to the function in question so the developer remembers to look at it.

controllers/subscriptions.go
jarrpa: If the IBM operator handles that, then we no. I'd rather let them decide how they manage their own components.

controllers/vendors.go
```diff
@@ -32,20 +32,28 @@ import (
...
+var FlashSystemKind = odfv1alpha1.StorageKind(strings.ToLower(reflect.TypeOf(ibmv1alpha1.FlashSystemCluster{}).Name()) +
+	"." + ibmv1alpha1.GroupVersion.String())
+
+var KnownKinds = []odfv1alpha1.StorageKind{
+	StorageClusterKind,
+	FlashSystemKind,
+}
+
 // VendorStorageCluster returns GroupVersionKind
 func VendorStorageCluster() odfv1alpha1.StorageKind {
```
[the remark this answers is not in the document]
jarrpa: I'd rather not. I did this on purpose because having these as global variables does the string generation only once, and no part of them relies on dynamic values at runtime. These functions were left here only because I didn't think it was worth it to completely remove them as it would have touched even more files than this PR already does. Ideally they'd also be `const`s, but go is the way it is.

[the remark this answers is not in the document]
jarrpa: After an offline discussion, agreed that we should clean this up for leave it for another PR.

Dockerfile
jarrpa: Done.

controllers/storagesystem_controller.go
jarrpa: Done.

controllers/storagesystem_controller.go
jarrpa: Got rid of it! Turns out it was just an artifact of a temporary solution. :joy:

Makefile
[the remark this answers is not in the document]
jarrpa: Done

iamniting (changes requested):

api/v1alpha1/storagesystem_types.go
```diff
@@ -27,7 +27,7 @@ import (
 
 //+kubebuilder:validation:Enum=flashsystemcluster.odf.ibm.com/v1alpha1;storagecluster.ocs.openshift.io/v1
 
-// StorageVendor captures the type of storage vendor
+// StorageKind specified a vendor's supported kinds
```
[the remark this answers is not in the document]
jarrpa: Dang it, of all things :joy:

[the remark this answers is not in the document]
jarrpa: Done!

iamniting (approved):
/lgtm

iamniting:
@jarrpa As expected it does require manual cherry-pick

jarrpa:
Of course! :joy:

PR #111, opened 2021-10-11
Fix creation and deletion of subscriptions

iamniting:
*What is achieved with the PR?*
* Prevent deletion of subscriptions while deleting the controller pod.
* Remove hack for creating subscriptions.
* Set odf-operator as the owner of owned subscriptions and CSV's.
* Fix isVendorCsvReady to check for all CSV's.
* Create ocs subscription on startup of the pod.
* Create a subscription controller.
* Wait for storageCluster CRD to be present before starting the controller.
* Do not run tests as part of the Dockerfile.

https://bugzilla.redhat.com/show_bug.cgi?id=2012718
https://bugzilla.redhat.com/show_bug.cgi?id=2014079
https://bugzilla.redhat.com/show_bug.cgi?id=2009531

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -254,6 +254,14 @@ spec:
           verbs:
           - update
         - apiGroups:
+          - operators.coreos.com
+          resources:
+          - subscriptions/status
+          verbs:
+          - get
+          - patch
+          - update
+        - apiGroups:
           - authentication.k8s.io
           resources:
           - tokenreviews
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -158,3 +158,11 @@ rules:
   - subscriptions/finalizers
   verbs:
   - update
+- apiGroups:
+  - operators.coreos.com
+  resources:
+  - subscriptions/status
+  verbs:
+  - get
+  - patch
+  - update


jarrpa:
> We do need to do get call because of the UID otherwise we see this error `Subscription.operators.coreos.com \"noobaa-operator\" is invalid: metadata.ownerReferences.uid: Invalid value: \"\": uid must not be empty`

Ah, okay. Still, you should only need to `Get()` once instead of doing it every loop. Just store the results in a higher-level variable or something, right?

controllers/subscriptions.go
```diff
@@ -73,95 +73,134 @@ func (r *StorageSystemReconciler) ensureSubscriptions(instance *odfv1alpha1.Stor
...
-		csvObj := &operatorv1alpha1.ClusterServiceVersion{}
-		err := r.Client.Get(context.TODO(), types.NamespacedName{
-			Name: csvName, Namespace: instance.Spec.Namespace}, csvObj)
+	var err error
 
+	csvObj := &operatorv1alpha1.ClusterServiceVersion{}
+	err = cli.Get(context.TODO(), types.NamespacedName{
+		Name: csvName, Namespace: OperatorNamespace}, csvObj)
+	if err != nil {
+		err = fmt.Errorf("CSV is not successfully installed")
```
iamniting, line 108: Let's say not found and add an existing error also instead of creating the new error

controllers/subscriptions.go
```diff
@@ -73,95 +73,134 @@ func (r *StorageSystemReconciler) ensureSubscriptions(instance *odfv1alpha1.Stor
...
+	})
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return nil, err
+	}
+
+	isReady := csvObj.Status.Phase == operatorv1alpha1.CSVPhaseSucceeded &&
+		csvObj.Status.Reason == operatorv1alpha1.CSVReasonInstallSuccessful
+
+	if !isReady {
+		err = fmt.Errorf("CSV is not successfully installed")
```
iamniting, line 123: maybe is not ready is better.

jarrpa: "Ready" isn't an actual thing as far as CSVs are concerned and is thus vague. We're just using it as shorthand internally.

controllers/subscription_controller.go
```diff
@@ -18,44 +18,209 @@ package controllers
...
+	}
 
 	return ctrl.Result{}, nil
 }
 
+func (r *SubscriptionReconciler) ensureSubscriptions(logger logr.Logger) error {
+
+	var err error
+
+	subsList := map[odfv1alpha1.StorageKind][]*operatorv1alpha1.Subscription{}
```
iamniting, line 80: We do not need to actually store the subscriptions we can just store the kind and get the subscription before calling `EnsureDesiredSubscription`. you can create a map of strings to boolean or strings to interface to reduce the map space also.

controllers/storagesystem_controller.go
```diff
@@ -217,39 +218,34 @@ func (r *StorageSystemReconciler) isVendorCsvReady(instance *odfv1alpha1.Storage
...
 
-	var errs []error
+	var returnErr error
 	for _, csvName := range csvNames {
 
 		csvObj, err := EnsureVendorCsv(r.Client, csvName)
 		if err != nil {
 			logger.Error(err, "failed to validate CSV", "ClusterServiceVersion", csvObj.Name)
-			errs = append(errs, err)
+			multierr.AppendInto(&returnErr, err)
```
iamniting, line 227: Can we introduce this change there itself where we wrote this code?

controllers/subscription_controller.go
jarrpa: Per our discussion, I used a map so I can just foo `map[blah] = foo` without having to do a `Contains()` or whatever. A map needs a key and its element. I use the Kind as the key for easy iteration through the StorageSystems, thus avoiding duplicates, and then use the same map to loop through the Subscription lists.

So, I'd rather leave this. :stuck_out_tongue:

PR #113, opened 2021-10-12
[HACK] pre-create OCS Subscriptions in StorageSystemReconciler.SetupWithManager()

jarrpa:
This PR pre-creates the OCS Subscription so as to re-enable the StorageClusterReconciler that was giving us CLBO after #99 was merged too hastily. The bug itself was that controller-runtime was expecting the StorageCluster CRD to exist *before* the controller manager is started, hence we are now creating the Subscription in `StorageSystemReconciler.SetupWithManager()`.

I really really REALLY REALLY don't like this, but I'm going through with it as a workaround for the installation and upgrade blockers in release-4.9. We'll need to take time to carefully consider a proper fix for this.

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/controllers/storagesystem_controller.go
+++ b/controllers/storagesystem_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -28,6 +29,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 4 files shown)


controllers/storagesystem_controller.go
```diff
@@ -209,6 +211,46 @@ func (r *StorageSystemReconciler) addReferenceToRelatedObjects(instance *odfv1al
 // SetupWithManager sets up the controller with the Manager.
 func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
+	// ========== START HACK ==========
+	// TODO: This is a hack to work around the absense of StorageCluster CRs.
```
obnoxxx, line 215: s/absense/absence/

controllers/storagesystem_controller.go
```diff
@@ -209,6 +211,46 @@ func (r *StorageSystemReconciler) addReferenceToRelatedObjects(instance *odfv1al
...
+	// A better fix should come in ASAP.
```
obnoxxx, line 216: Would be good to explain:
* why this fix is bad
* what a better fix would be
* what prevents us from doing the better fix and forces us do to the bad fix (for now)

obnoxxx (changes requested):
Reconsiler --> Reconciler

controllers/storagesystem_controller.go
obnoxxx: Yes, I am only requesting that this be explained somewhere. Commit message or PR description (or both).

jarrpa:
Remaining CI failure is expected, need to fix the test.

/override ci/prow/odf-operator-bundle-e2e-aws

PR #119, opened 2021-10-18
Add support for compatibility build

bipuladh:
Add support for OCP 4.10 + ODF 4.9 UI by alerting the basePath.
Basepath refers to the path where all the request are addressed to.

Signed-off-by: bipuladh <badhikar@redhat.com>

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -144,6 +144,38 @@ spec:
           - update
           - watch
         - apiGroups:
+          - apps
+          resources:
+          - deployments/finalizers
+          verbs:
+          - update
+        - apiGroups:
+          - config.openshift.io
+          resources:
+          - clusterversions
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
+          - config.openshift.io
+          resources:
+          - clusterversions/finalizers
+          verbs:
+          - update
+        - apiGroups:
+          - config.openshift.io
+          resources:
+          - clusterversions/status
+          verbs:
+          - get
+          - patch
+          - update
+        - apiGroups:
           - console.openshift.io
           resources:
           - consoleplugins
@@ -356,9 +388,7 @@ spec:
                 app: odf-console
             spec:
               containers:
-              - args:
-                - --ssl --cert=/var/serving-cert/tls.crt --key=/var/serving-cert/tls.key
-                image: quay.io/ocs-dev/odf-console:latest
+              - image: quay.io/ocs-dev/odf-console:latest
                 name: odf-console
                 ports:
                 - containerPort: 9001
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -41,6 +41,38 @@ rules:
   - update
   - watch
 - apiGroups:
+  - apps
+  resources:
+  - deployments/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - clusterversions
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - clusterversions/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - config.openshift.io
+  resources:
+  - clusterversions/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
   - console.openshift.io
   resources:
   - consoleplugins


console/console.go
```diff
@@ -95,7 +98,25 @@ func GetConsolePluginCR(pluginName string, displayName string, consolePort int,
...
+		return err
+	}
+	clusterVersion := ""
+	for _, version := range clusterVersionList.Items {
+		for _, history := range version.Status.History {
+			if history.State == configv1.CompletedUpdate {
+				clusterVersion = history.Version;
+			}
+		}
+	}
```
umangachapagain, line 114: I'd expect this to be already available in some openshift library. If not, we should probably have a dedicated `DetermineOpenShiftVersion()` function.

console/console.go
```diff
@@ -95,7 +98,25 @@ func GetConsolePluginCR(pluginName string, displayName string, consolePort int,
...
+	basePath := "release-4.9"
```
umangachapagain, line 115: add a newline above and remove newline from below.
Line 115 and 117 are logically same block.

console/console.go
```diff
@@ -95,7 +98,25 @@ func GetConsolePluginCR(pluginName string, displayName string, consolePort int,
...
+	// Determine ocp version
+	clusterVersionList := configv1.ClusterVersionList{}
+	if err := client.List(context.TODO(), &clusterVersionList); err != nil {
+		return err
+	}
+	clusterVersion := ""
+	for _, version := range clusterVersionList.Items {
+		for _, history := range version.Status.History {
+			if history.State == configv1.CompletedUpdate {
+				clusterVersion = history.Version;
```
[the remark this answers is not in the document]
bipuladh: there should only one occurrence. I am doing this to not rely on a particular name.

console/console.go
```diff
@@ -95,7 +98,16 @@ func GetConsolePluginCR(pluginName string, displayName string, consolePort int,
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=*
 
-func InitConsole(client client.Client, odfPort int) error {
+func InitConsole(client client.Client, odfPort int ) error {
+	// The base path to where the request are sent
+	basePath := ""
+	clusterVersion, err  := util.DetermineOpenShiftVersion(client)
```
jarrpa, line 104: I'll point out that this will only happen at operator startup. Thus, it seems entirely possible that the Console Pod itself would be on a version that is incompatible with our plugin until the odf-operator Pod is restarted. Is this the case?

pkg/util/openshift.go
```diff
@@ -0,0 +1,41 @@
...
+	clusterVersionList := configv1.ClusterVersionList{}
+	if err := client.List(context.TODO(), &clusterVersionList); err != nil {
+		return "", err
+	}
+	clusterVersion := ""
+	for _, version := range clusterVersionList.Items {
+		for _, history := range version.Status.History {
+			if history.State == configv1.CompletedUpdate {
+				clusterVersion = history.Version;
+			}
```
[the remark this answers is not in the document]
bipuladh: Looks like op-framework also relies on CV. I can start using `status.desired.version` like they use.

pkg/util/openshift.go
```diff
@@ -0,0 +1,37 @@
...
+func DetermineOpenShiftVersion (client client.Client) (string, error) {
+	// Determine ocp version
+	clusterVersionList := configv1.ClusterVersionList{}
+	if err := client.List(context.TODO(), &clusterVersionList); err != nil {
+		return "", err
+	}
+	clusterVersion := ""
+	for _, version := range clusterVersionList.Items {
+		clusterVersion = version.Status.Desired.Version
+	}
```
iamniting, line 35: What if this list is empty and this will send back "",nil which is not expected I guess. maybe one more check to check a empty value is required

controllers/clusterversion_controller.go
```diff
@@ -18,44 +18,86 @@ package controllers
...
+	}
 
 	return ctrl.Result{}, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
+
+//+kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=*
```
iamniting, line 66: These markers should be on top of Reconcile

console/console.go
```diff
@@ -32,6 +32,8 @@ import (
 )
 
 const DEPLOYMENT_NAMESPACE = "openshift-storage"
+const MAIN_BASE_PATH = ""
```
iamniting, line 35: lets make use of these variables, I do not a use of them

iamniting, line 35: These changes should be also in the first commit

jarrpa: Disagreed. The first commit should introduce no new code that wasn't the result of running the initialization command. Otherwise we won't easily know what was generated and what wasn't.

pkg/util/openshift.go
jarrpa: This list *shouldn't* ever be empty. If it is, then we have something dreadfully wrong. Maybe we should consider returning a more verbose message about this problem, but as it stands this is fine for now until if and when we ever run into this.

PR #136, opened 2021-10-28
makefile: remove vendor-update target

iamniting:
running `go mod vendor` while running tests in the container downloads
the all deps again eventhoug they are already installed. Moreover go
already do these checks and verify go.mod with the vendor/modules.txt.

The example runs show this target is not required:

https://github.com/iamniting/odf-operator/runs/4035819892?check_suite_focus=true
https://github.com/iamniting/odf-operator/runs/4035819310?check_suite_focus=true

It does save the unnecessary download twice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,10 +43,10 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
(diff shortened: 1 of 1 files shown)


jarrpa (approved):
/lgtm

PR #139, opened 2021-10-29
Mirror golang images to ocs-dev and start using them as a base

iamniting:

### DIFF
--- a/.github/workflows/mirror-noobaa-upstream-images.yaml
+++ b/.github/workflows/mirror-noobaa-upstream-images.yaml
@@ -1,5 +1,5 @@
 ---
-name: mirror-noobaa-upstream-images
+name: mirror-upstream-images
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


jarrpa (changes requested):
I think this is not needed anymore? With the following lines in the OCP CI we shouldn't be trying to pull from Docker Hub anymore: https://github.com/openshift/release/blob/master/ci-operator/config/red-hat-storage/odf-operator/red-hat-storage-odf-operator-main.yaml#L19-L22

iamniting:
> I think this is not needed anymore? With the following lines in the OCP CI we shouldn't be trying to pull from Docker Hub anymore: https://github.com/openshift/release/blob/master/ci-operator/config/red-hat-storage/odf-operator/red-hat-storage-odf-operator-main.yaml#L19-L22

I think so. We can always open it if required in the future.

PR #144, opened 2021-11-11
console: Fix basepath for odf-console deployment

vbnrh:
There were a couple of changes introduced in OCP 4.10 which caused the
dynamic plugin to break. 

Signed-off-by: Vineet Badrinath <vineetbnath@gmail.com>

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -26,8 +26,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-const MAIN_BASE_PATH = ""
-const COMPATIBILITY_BASE_PATH = "compatibility"
+const MAIN_BASE_PATH = "/"
+const COMPATIBILITY_BASE_PATH = "/compatibility/"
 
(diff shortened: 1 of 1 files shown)


console/console.go
```diff
@@ -26,8 +26,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-const MAIN_BASE_PATH = ""
-const COMPATIBILITY_BASE_PATH = "compatibility"
+const MAIN_BASE_PATH = "/"
+const COMPATIBILITY_BASE_PATH = "/compatibility"
```
bipuladh, line 30:
```suggestion
const COMPATIBILITY_BASE_PATH = "/compatibility/"
```

bipuladh (changes requested):

console/console.go
vbnrh: done

iamniting (approved):
/lgtm

PR #151, opened 2021-11-22
controllers: add StorageClass encryption page to ODF QuickStart guide

SanjalKatiyar:
adds one more step "Encrypt your StorageClass" to "Configure and manage OpenShift Data Foundation" QuickStart guide.
Screenshot from 2021-11-22 19-54-38
Screenshot from 2021-11-22 19-54-51

Epic URL: https://issues.redhat.com/browse/RHSTOR-1950
Signed-off-by: sanjalkatiyar <sanjaldhir@gmail.com>

### DIFF
--- a/controllers/quickstart_constants.go
+++ b/controllers/quickstart_constants.go
@@ -238,6 +238,54 @@ spec:
         failedTaskHelp: This task isn’t verified yet. Try the task again.
       summary:
         success: You have successfully created BucketClass
+        failed: Try the steps again. 
+    - title: Encrypt your StorageClass
+      description: |-
+
+          StorageClass encryption enables you to provide individual keys to each of your StorageClusters.
+
+
+          You can create an encrypted StorageClass while configuring your StorageCluster, or when creating a new StorageClass.
+
+
+          **To create an encrypted StorageClass while setting up your cluster:**
+
+          1. In the navigation menu, select [Operators]{{highlight qs-nav-operators}} and select **Installed Operators**, then select **Create StorageSystem**.
+
+          2. In the creation wizard, go to **Security and network** step.
+
+          3. Check the **StorageClass encryption** box.
+
+          4. Select your **Key management service provider**.
+
+          5. Provide your connection details.
+
+
+          **To create a new encrypted StorageClass:**
+
+          1. In the navigation menu, select [Storage]{{highlight qs-nav-storage}} and select **StorageClasses**, then select **Create StorageClass**.
+
+          2. In the **Provisioner** field, select **openshift-storage.rbd.csi.ceph.com**.
+
+          3. Check the **Enable Encryption** box.
+
+          4. Select your **Key management service provider**.
+
+          5. Provide your connection details.
+      review:
+        instructions: |-
+          ####  To verify that the StorageClass is encrypted:
+          1. Go to **Storage > StorageClasses**.
+          If you encrypted the StorageClass while setting up your cluster, check that a new **ocs-storagecluster-ceph-rbd-encrypted** StorageClass is created.
+
+          2. Select the newly created StorageClass and select the **YAML** tab.
+
+          Does the value of the parameter **encrypted** equal to **true**?
+
+          Is the value of the parameter **encryptionKMSID** equal to the KMS connection's service name?
+        failedTaskHelp: This task isn’t verified yet. Try the task again.
+      summary:
+        success: You have successfully encrypted StorageClass
         failed: Try the steps again.
 
   conclusion: You're ready to go! Now you can customize your StorageSystems in OpenShift Data Foundation.


iamniting:
@SanjalKatiyar Do we need to backport it? If yes can we have a bug pls?

SanjalKatiyar:
> @SanjalKatiyar Do we need to backport it? If yes can we have a bug pls?

This is a 4.10 epic (https://issues.redhat.com/browse/RHSTOR-1950), so, we don't need to backport it.

controllers/quickstart_constants.go
```diff
@@ -238,6 +238,54 @@ spec:
...
+
+          **To create a new encrypted StorageClass:**
+
+          1. In the navigation menu, select **Storage > StorageClasses**, and select **Create StorageClass**.
+
+          2. In the **Provisioner** field, select **openshift-storage.rbd.csi.ceph.com**.
+
+          3. Check the **Enable Encryption** box.
+
+          4. Select your **Key management service provider**.
```
vbnrh, line 272:
```suggestion
          4. Select your **KMS provider**.
```

controllers/quickstart_constants.go
```diff
@@ -238,6 +238,54 @@ spec:
...
+
+          StorageClass encryption enables you to provide individual keys to each of your StorageClusters.
+
+
+          You can create an encrypted StorageClass while configuring your StorageCluster, or when creating a new StorageClass.
+
+
+          **To create an encrypted StorageClass while setting up your cluster:**
+
+          1. After installing OpenShift Data Foundation, select **Create StorageSystem**.
```
vbnrh, line 253: Path to Openshift Data Foundation ? We should probably guide them to the operators

vbnrh: In the navigation menu, select **Storage > OpenShift Data Foundation**.

controllers/quickstart_constants.go
```diff
@@ -238,6 +238,54 @@ spec:
...
+
+          3. Check the **Enable Encryption** box.
+
+          4. Select your **Key management service provider**.
+
+          5. Provide your connection details.
+      review:
+        instructions: |-
+          ####  To verify that the StorageClass is encrypted:
+          1. Go to **Storage > StorageClasses**.
```
vbnrh, line 278: We can probably use the perspective highlighters to point to menu in the side navigation.

controllers/quickstart_constants.go
```diff
@@ -238,6 +238,54 @@ spec:
...
+          If you encrypted the StorageClass while setting up your cluster, check that a new StorageClass with the prefix/suffix **encrypted** is created.
```
vbnrh, line 279:
```suggestion
          If you encrypted the StorageClass while setting up your cluster, check that a new StorageClass with the prefix/suffix **encrypted** is created.
```

How can it be both prefix or suffix ? Is it dependent on user to choose that ?

controllers/quickstart_constants.go
```diff
@@ -238,6 +238,54 @@ spec:
...
+
+          2. Select the newly created StorageClass and select the **YAML** tab.
+
+          Does the value of the parameter **encrypted** equal to **true**?
+
+          Is the value of the parameter **encryptionKMSID** equal to the KMS connection's **Service name**?
```
vbnrh, line 285:
```suggestion
          Is the value of the parameter **encryptionKMSID** equal to the KMS connection's service name?
```
No need to bold generic terms unless it is a field name, in which case use the proper field name

vbnrh (changes requested):
Pretty good. We can use perspective highlighters to point to navigation items

controllers/quickstart_constants.go
```diff
@@ -238,6 +238,54 @@ spec:
...
+
+          **To create an encrypted StorageClass while setting up your cluster:**
+
+          1. After installing OpenShift Data Foundation, select **Create StorageSystem**.
+
+          2. In the creation wizard, go to **Security and network** step.
+
+          3. Check the **StorageClass encryption** box.
+
+          4. Select your **Key management service provider**.
```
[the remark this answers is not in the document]
SanjalKatiyar: `Key management service provider` is the field name which we show in the UI.

controllers/quickstart_constants.go
SanjalKatiyar: same as above... `Key management service provider` is the exact name that we show in the UI.

controllers/quickstart_constants.go
SanjalKatiyar: It will be either one. It is not dependent on user, it is something OCS operator will create. It is yet to be implemented in 4.10

controllers/quickstart_constants.go
SanjalKatiyar: `Service name` is the proper field name that we show in the UI.

vbnrh: but we are in the YAML tab i presume after step 2 ?
`Select the newly created StorageClass and select the **YAML** tab.`
in which case we can use the field name or keep it as `service name` itself

vbnrh (changes requested):

controllers/quickstart_constants.go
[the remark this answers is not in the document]
vbnrh: I think we don't need to use the entire name. Abbreviation should work well here. We can assume the admin would know what KMS is if they are encrypting something.
wdyt ?

controllers/quickstart_constants.go
SanjalKatiyar: We have already documented these steps in "Install OpenShift Data Foundation" guide. Adding it here as well will unnecessarily add extra steps to the page.
We will have to add these steps again, if we want to tell user how to create a StorageSystem which IMO is might not be required. Plz let me know your thoughts.
1) In the main navigation menu, select  and select Installed Operators.
2) From your list of installed operators, select OpenShift Data Foundation.
3) On the StorageSystem tile, select Create Instance.

controllers/quickstart_constants.go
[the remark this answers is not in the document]
SanjalKatiyar: U r definitely right, but, we should have consistence with other guide/pages as well. Everywhere we are using bold text we are using exactly same text which is shown in the UI. So, let's keep it that way everywhere.

controllers/quickstart_constants.go
SanjalKatiyar: This need more discussion. In the other pages of this Configure and Manage guide, we are using same format, i.e.,
**Storage > StorageClasses** (no perspective highlighters). So, for this PR we should have it same way as other pages. May be we can make these changes (if required) later on in all the pages in one go, let this PR be specific to the epic related changes.
WDYT ??

controllers/quickstart_constants.go
SanjalKatiyar: ack... I will change it to 
```suggestion
          Is the value of the parameter **encryptionKMSID** equal to the KMS connection's service name?
```

controllers/quickstart_constants.go
SanjalKatiyar: as per discussion, added it...

iamniting (approved):
/lgtm

PR #152, opened 2021-12-10
Update storageCluster API

iamniting:
Signed-off-by: Nitin Goyal nigoyal@redhat.com

https://bugzilla.redhat.com/show_bug.cgi?id=2031023

Depends on https://github.com/red-hat-storage/odf-operator/pull/153


jarrpa:
If we are officially dropping support for go 1.15, we should update the CI to reflect that. Probably push another PR to disable it and then we can waive it here.

jarrpa:
Changes look good... but it seems that a change in how controller-runtime is doing things caused our unit tests to break in a way that very few would run into. So of course we did! The following PR in the CI should help alleviate this: https://github.com/openshift/release/pull/24507

jarrpa:
Success!

/lgtm

PR #170, opened 2022-01-21
controllers: start filtering odf sub instead of using static name

iamniting:
While creating a odf-operator subscription from the UI it is always
created by the name of odf-operator. Now managed services is gonna use
the odf-operator as a dependency which means the name will be OLM
generated name and we can not assume static name anymore in the
managed cluster.

Filtering out is a solution to the problem.


jarrpa (approved):
/lgtm

PR #172, opened 2022-01-25
controllers: change controller ref to owner ref of the vendor system

bindrad:
We are using the controller ref to show the vendor CR in the OCP UI and can also use the owner ref for the same.
controller ref is required by the managed services as storageCluster is created by them in the manages services.

changing the controller ref to owner ref will solve the managed services problem.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2044447

Signed-off-by: bindrad <dbindra@redhat.com>


iamniting (changes requested):
Hi Dhruv, Changes looks good to me, CI is also passed, Can we pls change a commit-msg with something descriptive. eg.

```
controllers: change controller ref to owner ref of the vendor system

We are using the controller ref to show the vendor CR in the OCP UI
and can also use the owner ref for the same.
controller ref is required by the managed services as storageCluster 
is created by them in the manages services.

changing the controller ref to owner ref will solve the managed 
services problem.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2044447
```

iamniting (approved):
/lgtm

agarwal-mudit:
No need to cherry-pick, change is already there in release-4.10

PR #178, opened 2022-02-23
Use a unique label for monitoring resources

afreen23:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2043034

Issue
=====
Both csi-addon operator and odf-operator are using the default label and label-selector provided by operator-sdk - "control-plane: controller-manager" for its services and deployments.
This is allowing deployments, services and service monitors of each operator to select the pods ans services of other operator.
One critical impact of this is in UI monitoring where odf metrics monitor is picking up csiaddon service and hence showing an extra storage system. This is breaking the prometheus query and hence missing storage system on UI.

Fix
===
Use a unique label selector in odf-operator to avoid conflicts.

Before:
Screenshot from 2022-02-23 21-28-06

After:
Screenshot from 2022-02-23 19-44-12

### DIFF
--- a/bundle/manifests/odf-operator-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/odf-operator-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -2,7 +2,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
-    control-plane: controller-manager
+    app.kubernetes.io/name: odf-operator
   name: odf-operator-controller-manager-metrics-monitor
 spec:
   endpoints:
@@ -14,4 +14,4 @@ spec:
       insecureSkipVerify: true
   selector:
     matchLabels:
-      control-plane: controller-manager
+      app.kubernetes.io/name: odf-operator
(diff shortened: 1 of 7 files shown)


afreen23:
@iamniting @umangachapagain please review

iamniting:
@afreen23 changes look good to me as they are really simple, But I have questions before I give LGTM. 

1. Will this require a change in the openshift 4.10 console code too.
2. If yes Will openshift console code work with the odf 4.9 until we release 4.10 as openshift gets released before odf.

afreen23:
> @afreen23 changes look good to me as they are really simple, But I have questions before I give LGTM.
> 
> 1. Will this require a change in the openshift 4.10 console code too.
> 2. If yes Will openshift console code work with the odf 4.9 until we release 4.10 as openshift gets released before odf.
> 3. or this has nothing to do with the openshift console code and matches labels b/w themselves.

Thanks for the review @iamniting 
No change is required in openshift 4.10 console code . 

**Details:**
The [prometheus query](https://github.com/red-hat-storage/odf-console/blob/34711cbcf7c7327fb2fa66b3287d2d1da7caa18e/packages/odf/components/odf-dashboard/queries.ts#L29) used on dashboard is failing currently. odf_system_map query is listing two storage systems (due to https://github.com/red-hat-storage/odf-operator/pull/178#issue-1148027973) and when we join it with another query it starts failing since many to many matching occurs.

Screenshot from 2022-02-24 00-35-25

PR #184, opened 2022-02-28
controllers: add ocs toleration to csi-addons subscription

Rakshith-R:
This commit adds the following tolerationt to
csi-addons subscription so it is applied to
all the pods it creates.
```
      - effect: NoSchedule
        key: node.ocs.openshift.io/storage
        operator: Equal
        value: "true"
```

Currently all the pods related to storage is expected to have the
above toleration to be able to tolerate taints on node dedicated for storage(refer [1]).
The other operators have this in their manager.yaml files. 
By adding this to csiaddons subscription(which will pass it onto any pods created
by it, refer [2}), we can avoid having it in upstream code.

Signed-off-by: Rakshith R <rar@redhat.com>

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2059105
[2] https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md#tolerations

### DIFF
--- a/controllers/subscriptions.go
+++ b/controllers/subscriptions.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"go.uber.org/multierr"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


yati1998 (approved):

nixpanic:
The commit message (or PR description) does not explain why this is needed, or what the expected result of the toleration should be. Without a description of the intended behaviour, I do not feel comfortable approving this.

Rakshith-R:
> The commit message (or PR description) does not explain why this is needed, or what the expected result of the toleration should be. Without a description of the intended behaviour, I do not feel comfortable approving this.

Added description of the problem,fix and references, PTAL 
/cc @jarrpa

jarrpa (approved):
/lgtm

PR #189, opened 2022-03-01
controllers: stop adding annotation to the storagecluster

iamniting:
adding an annotation to the storagecluster changes the resource version
of the storagecluster on etcd and ocs-operator fails to update the
status while onboarding the consumer.

There is no easy way to get the id of the storage consumer from the
provider again after the onboarding is successful from the provider end.

minimize the update calls wherever possible.

https://bugzilla.redhat.com/show_bug.cgi?id=2056634


umangachapagain (approved):
ocs-operator could fail status updates for many reasons. Protection should be built into ocs-operator for that.

If StorageSystem can be created for all StorageCluster without annotations and using predicates, we can drop annotations. Not because it might prevent status update failures in ocs-operator, but because we have better solutions available here.

lgtm

PR #193, opened 2022-03-28
update go to 1.17 and update csv version to 4.11.0

iamniting:
- Update golang to 1.17 and its deps
- Update all actions to make them compatible with golang 1.17
- Update CSV version
- Use go install instead of depricated go get


.github/workflows/codespell.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
       - name: codespell
         uses: codespell-project/actions-codespell@master
         with:
+          exclude_file: go.sum
```
umangachapagain, line 20: Why do we need to exclude it?

umangachapagain:
@iamniting There are multiple changes in this PR (not just Go 1.17 change).
I suggest updating PR title and description to make it clearer.

.github/workflows/codespell.yaml
iamniting: It was failing here https://github.com/red-hat-storage/odf-operator/actions/runs/2050857294

iamniting:
> @iamniting There are multiple changes in this PR (not just Go 1.17 change). I suggest updating PR title and description to make it clearer.

done

jarrpa (approved):
/lgtm

PR #198, opened 2022-04-07
End to end test: Added test suite functions

malayparida2000:
This test installs the odf operator & starts a storage system. Then it unistalls the operator & cleans up everything.


e2e/setup_teardown.go
```diff
@@ -0,0 +1,40 @@
+package e2e
+
+import (
+	"github.com/onsi/gomega"
+)
+
+// BeforeTestSuiteSetup is the function called to initialize the test environment
+func BeforeTestSuiteSetup() {
```
iamniting, line 8: lets rename it to Setup()

e2e/setup_teardown.go
```diff
@@ -0,0 +1,40 @@
...
+
+	gomega.Expect(err).To(gomega.BeNil())
+
+	SuiteFailed = false
+
+	debug("------------------------------\n")
+}
+
+// AfterTestSuiteCleanup is the function called to tear down the test environment
+func AfterTestSuiteCleanup() {
```
iamniting, line 28: rename this one to TearDown()

pkg/deploymanager/ODFStorageSystem.go
```diff
@@ -0,0 +1,34 @@
+package deploymanager
```
iamniting, line 1: rename the files using underscore with all small case

pkg/deploymanager/config.go
```diff
@@ -0,0 +1,133 @@
...
+// InstallNamespace is the namespace odf is installed into
+const InstallNamespace = "openshift-storage"
+
+//nolint:errcheck // ignoring err check as causing failures
+func init() {
+	odfv1alpha1.SchemeBuilder.AddToScheme(scheme.Scheme)
+}
+
+type DeployManager struct {
+	olmClient      *olmclient.Clientset
```
iamniting, line 29: lets use one client for all of the schemes.

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,311 @@
...
+	odfOG := v1.OperatorGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "openshift-storage-operatorgroup",
+			Namespace: InstallNamespace,
+		},
+		Spec: v1.OperatorGroupSpec{
+			TargetNamespaces: []string{InstallNamespace},
+		},
+	}
+	odfOG.SetGroupVersionKind(schema.GroupVersionKind{Group: v1.SchemeGroupVersion.Group, Kind: "OperatorGroup", Version: v1.SchemeGroupVersion.Version})
```
iamniting, line 73: I don't think we should do this

iamniting (changes requested):

pkg/deploymanager/config.go
malayparida2000: Having some doubts & issues here as these clients are used separately inside the subscription.go file.

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,322 @@
...
+	label := make(map[string]string)
+	// Label required for monitoring this namespace
+	label["openshift.io/cluster-monitoring"] = "true"
+
+	// Namespaces
+	co.namespaces = append(co.namespaces, k8sv1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Namespace",
+			APIVersion: "v1",
+		},
```
iamniting, line 55: I don't think this is required.

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,322 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   InstallNamespace,
+			Labels: label,
```
iamniting, line 58: Add labels directly instead of using variables.

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,322 @@
...
+
+func (t *DeployManager) generateClusterObjects(odfCatalogImage string, subscriptionChannel string) *clusterObjects {
+
+	co := &clusterObjects{}
+	label := make(map[string]string)
+	// Label required for monitoring this namespace
+	label["openshift.io/cluster-monitoring"] = "true"
+
+	// Namespaces
+	co.namespaces = append(co.namespaces, k8sv1.Namespace{
```
iamniting, line 51: create a variable and append it later, as you are doing for other variables.

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,322 @@
...
+	co.subscriptions = append(co.subscriptions, odfSubscription)
+
+	return co
+}
+
+// Deploying the cluster objects
+func (t *DeployManager) deployClusterObjects(co *clusterObjects) error {
+
+	for _, namespace := range co.namespaces {
+		err := t.CreateNamespace(namespace.Name)
```
iamniting, line 130: As you are just using a name, why not to create a list of namespaces instead of creating a object.

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,322 @@
...
+		},
+		Spec: v1alpha1.CatalogSourceSpec{
+			SourceType:  v1alpha1.SourceTypeGrpc,
+			Image:       odfCatalogImage,
+			DisplayName: "OpenShift Data Foundation",
+			Publisher:   "Red Hat",
+		},
+	}
+
+	co.catalogSources = append(co.catalogSources, odfCatalog)
```
iamniting, line 104: Why do we need catalog twice?

pkg/deploymanager/subscription.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+	}
+
+	for _, subscription := range co.subscriptions {
+		err := t.olmClient.OperatorsV1alpha1().Subscriptions(subscription.Namespace).Delete(context.TODO(), subscription.Name, metav1.DeleteOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
```
iamniting, line 320: This order should be a exact opposite of create

iamniting (changes requested):

pkg/deploymanager/subscription.go
malayparida2000: I think the clusterobject here has the namespace, operatorgroup, catalogsource & subscription. So it can not be a list I think

iamniting: It can be, lets discuss this over a call

iamniting (approved):

PR #200, opened 2022-04-12
Added a "object storage" description about Ceph

hector-vido:
Sometimes people think that only with Noobaa they can create `ObjectBucketClaims`.
Probably they forget or don't know about the **ocs-external-storagecluster-ceph-rgw** `StorageClass`.
I added in the Ceph description the "object" volume type.
This is pretty simple, maybe don't do any difference at all but is good to tell about this.
From time to time I got someone thinking in this way.
Thank you!

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -93,8 +93,8 @@ spec:
     scope as well as watching and configuring their CustomResources.
 
     ### Rook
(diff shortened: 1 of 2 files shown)


iamniting:
Hello @hector-vido Thanks for the PR, Can you pls follow our Contribution guide [1] to write a commit message as per our requirements. Also pls sign off the commit.

[1]. https://github.com/red-hat-storage/odf-operator/blob/main/CONTRIBUTING.md#commit-structure

iamniting (approved):
/lgtm

PR #202, opened 2022-04-15
test: Fix WaitForCsv fucntion in the deploymanager package

malayparida2000:
After the subscriptions are posted it takes some time for the Csvs to
come up. So in that time we shoudn't proceed with an empty csvlist.
This commit fixes that by checking for the csvlist is empty or not
before starting the wait.


iamniting (changes requested):
We should not do this change. This change does not make sure that we are installing the correct CSV, Instead, we need to remove the list method and check via specific names.

malayparida2000:
Creating a new pr with the new approach for fixing this problem

PR #204, opened 2022-04-18
godeps: Remove go install called on gomega

malayparida2000:
Gomega is just a matcher library and it's already in the
go.mod of our project. It need not be installed like the
ginkgo binary. Calling go install on gomega gives a
warning of matched no package.

Signedoff-by: Malay Kumar Parida <mparida@redhat.com>


iamniting (changes requested):
Can you change the type to `makefile`?

iamniting (approved):
/lgtm

PR #205, opened 2022-04-18
godeps: update ginkgo & gomega version

malayparida2000:
With the release of Ginkgo 2.0 the 1.x version is formally
deprecated and no longer supported. So upgrading ginkgo to
v2 from v1. Also updating gomega to 1.19.


iamniting (changes requested):
can you change the type of the second commit from `godeps` to `test`? Also, split the third commit into 2 one for go.mod and go.sum and another one for vendor changes.

iamniting (approved):
/lgtm

PR #207, opened 2022-04-25
makefile: update make-tools and remove unnecessary code

iamniting:
while using go get we were always creating a new temp directory and go.mod. So go do not make any changes to the existing go.mod while downloading the binaries. This is not required anymore as go install does not make changes to the existing go.mod if we give the full path with version.

ref: https://go.dev/doc/go-get-install-deprecation

### DIFF
--- a/hack/make-tools.mk
+++ b/hack/make-tools.mk
@@ -1,13 +1,8 @@
 # go-get-tool will 'go get' any package $2 and install it to $1.
 define go-get-tool
 @[ -f $(1) ] || { \
-set -e ;\
-TMP_DIR=$$(mktemp -d) ;\
-cd $$TMP_DIR ;\
-go mod init tmp ;\
 echo "Downloading $(2)" ;\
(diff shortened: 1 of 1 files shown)


hack/make-tools.mk
```diff
@@ -1,13 +1,8 @@
 # go-get-tool will 'go get' any package $2 and install it to $1.
 define go-get-tool
 @[ -f $(1) ] || { \
-set -e ;\
-TMP_DIR=$$(mktemp -d) ;\
-cd $$TMP_DIR ;\
-go mod init tmp ;\
```
jarrpa, line 7: I never quite understood why the scripts were generated this way. If you can provide some description as to what they were trying to do and why we don't need it, that would make me feel better about this.

iamniting: added it in the description.

jarrpa (approved):
/lgtm

PR #208, opened 2022-04-26
test: Check for all the required CSV names in the SetUp function in the BeforeSuite

malayparida2000:
We check for all the CSVs, if they have reached succeeded state or not. This is done after the operator installation is completed in the SetUp Function in the BeforeSuite

### DIFF
--- a/pkg/deploymanager/olm.go
+++ b/pkg/deploymanager/olm.go
@@ -35,6 +35,23 @@ func (d *DeployManager) DeployODFWithOLM(odfCatalogImage, subscriptionChannel, o
 	return nil
 }
 
+// CheckAllCsvs checks if all the required csvs are present & have succeeded
+func (d *DeployManager) CheckAllCsvs(csvNames []string) error {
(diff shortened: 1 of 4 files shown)


pkg/deploymanager/olm.go
```diff
@@ -129,16 +146,6 @@ func (d *DeployManager) CreateOlmResources(olmResources *OlmResources) error {
...
-		csv := &operatorsv1alpha1.ClusterServiceVersion{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      subscription.Spec.StartingCSV,
-				Namespace: subscription.Namespace,
-			},
-		}
-		err = d.WaitForCsv(csv)
-		if err != nil {
-			return err
-		}
```
malayparida2000, line 141: Have removed this from here as after the operator is installed, In the end, the CheckAllCsvs function will check for all the csvs have succeeded or not. So this is no longer necessary.

iamniting (approved):
/lgtm

PR #212, opened 2022-05-11
Bug 2064314: console: update basePath for ConsolePlugin

bipuladh:
BasePath needs to be set as per the version of the OCP cluster

Signed-off-by: Bipul Adhikari <badhikar@redhat.com>

### DIFF
--- a/controllers/clusterversion_controller.go
+++ b/controllers/clusterversion_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


bipuladh:
BP required till 4.9

controllers/clusterversion_controller.go
```diff
@@ -107,8 +109,10 @@ func (r *ClusterVersionReconciler) ensureConsolePlugin(clusterVersion string) er
 	}
 
 	// Create/Update ODF console ConsolePlugin
-	odfConsolePlugin := console.GetConsolePluginCR(r.ConsolePort, basePath, OperatorNamespace)
+	odfConsolePlugin := console.GetConsolePluginCR(r.ConsolePort, OperatorNamespace)
 	_, err = controllerutil.CreateOrUpdate(context.TODO(), r.Client, odfConsolePlugin, func() error {
+		logger.Info(fmt.Sprintf("Set the BasePath for odf-console plugin as '%s'", basePath))
```
jarrpa, line 114: I don't like this log message here, since this will lead to potentially massive spam in our logs. If you really want to capture this event, first check to see if the current basePath value differs from our expected value, and only then output the log message.

jarrpa (changes requested):
Overall looks good, but one comment on the logging strategy here.

PR #215, opened 2022-05-18
controllers: update the status before adding/removing finalizers

iamniting:
update the status before the Update call as update call does not
update the status and we lost the status.

Which causes some missing conditions.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2086423


iamniting:
@jarrpa The another way to fix the same is [here](https://github.com/iamniting/odf-operator/commit/81d3fc25954fff3357c714e220ac52b041c5cab8)

Let me know which one you like?

iamniting:
closing this in favour of https://github.com/red-hat-storage/odf-operator/pull/218

PR #220, opened 2022-05-30
test: Correct logical mistake in fixing panic happening in WaitForCatalogsource

malayparida2000:
A Panic was happening in WaitForCatalogsource during runtime. So a
null check was added to It. But while doing so a logical mistake was
left behind which allowed it to return true even when the connection
state was nill. Now fixing it.


iamniting (approved):
/lgtm

PR #221, opened 2022-05-30
bundle: Add csi-addons to the dependencies.yaml

iamniting:
csi-addons resolution gets fail when user uses the custom catalog
source as it is not specified in the dependencies.yaml and the sub is
not created via OLM. The sub is created by the odf directly which does
not have a knowledge about the custom catalog source and it is still
pointing to the redhat-operators in the openshift-marketplace namespace.

Managing there own catalog source is seen very offently now a days as it
is MS and disconnected ENV usecase.

https://bugzilla.redhat.com/show_bug.cgi?id=2056697

### DIFF
--- a/bundle/metadata/dependencies.yaml
+++ b/bundle/metadata/dependencies.yaml
@@ -3,3 +3,7 @@ dependencies:
   value:
     packageName: ocs-operator
     version: ">=4.10.0 <4.11.0 || 4.11.0"
+- type: olm.package
+  value:
+    packageName: csi-addons
+    version: "0.2.0"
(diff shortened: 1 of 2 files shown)


bundle/metadata/dependencies.yaml
```diff
@@ -3,3 +3,7 @@ dependencies:
   value:
     packageName: ocs-operator
     version: ">=4.10.0 <4.11.0 || 4.11.0"
+- type: olm.package
+  value:
+    packageName: csi-addons
+    version: "0.2.0"
```
nb-ohad, line 9: Don't you need the fancy `>=`, `<`, `||` pattern to ensure an upgrade to a new y stream?

iamniting: Yes, we need it downstream, Boris will change those values in the Downstream build.

nb-ohad: Ok

jarrpa: Since we're currently not testing upgrades in our CI, this isn't a technical problem for acceptance upstream. We should fix that, of course. Later.

PR #222, opened 2022-05-31
[wip] test: Add an Upgrade test into the e2e test

malayparida2000:
This adds an upgrade test to the existing e2d testing in odf. We install an old version of odf and upgrade from it to the latest one.
After upgrading we check if all the csvs have succeded & whether the storagesystem is still healthy.


malayparida2000:
There is no plan to work on this in near future, so closing this PR

PR #225, opened 2022-06-09
bundle: use extra label for odf-operator configs

malayparida2000:
The odf-operator-controller-manager-metrics-service uses only one label
control-plane: controller-manager as pod selector. A different pod also
uses the same label, the service does select the other pod as well along
with our pod and creates issues.

Add the extra label to solve this conflict.

The extra label has been added to all the places except the selector of
odf-operator-controller-manager deployment. As adding the extra label
there causes upgrade failure.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2089342


iamniting (approved):
/lgtm

PR #233, opened 2022-07-26
Several pent up frustrations ;)

jarrpa:
These are a bulk of patches I've been maintaining in my working branch for a while. I've been using them regularly, so at least they WFM and I think I've done an all right job of keeping them out of the way for anyone who doesn't want them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
+lint: ## Run golangci-lint against code.
+	docker run --rm -v $(PROJECT_DIR):/app -w /app $(GO_LINT_IMG) golangci-lint run -E gosec --timeout=6m .
+
 godeps-update: ## Run go mod tidy and go mod vendor.
 	go mod tidy && go mod vendor
 
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,7 @@ bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metada
 	cd config/default && $(KUSTOMIZE) edit set image rbac-proxy=$(RBAC_PROXY_IMG)
 	cd config/console && $(KUSTOMIZE) edit set image odf-console=$(ODF_CONSOLE_IMG)
 	cd config/manifests/bases && $(KUSTOMIZE) edit add annotation --force 'olm.skipRange':"$(SKIP_RANGE)" && \
+	        $(KUSTOMIZE) edit add annotation --force 'operators.operatorframework.io/operator-type':"$(OPERATOR_TYPE)" && \
 		$(KUSTOMIZE) edit add patch --name odf-operator.v0.0.0 --kind ClusterServiceVersion\
 		--patch '[{"op": "replace", "path": "/spec/replaces", "value": "$(REPLACES)"}]'
 	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
--- a/Makefile
+++ b/Makefile
@@ -157,8 +157,8 @@ bundle-push: ## Push the bundle image.
 # This recipe invokes 'opm' in 'semver' bundle add mode. For more information on add modes, see:
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-build
-catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+catalog-build: ## Build a catalog image.
+	docker build --build-arg BUNDLE_IMGS=${BUNDLE_IMGS} -f index.Dockerfile -t $(CATALOG_IMG) .
 
 # Push the catalog image.
 .PHONY: catalog-push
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -135,13 +135,16 @@ deploy-with-olm: kustomize ## Deploy controller to the K8s cluster via OLM
 undeploy-with-olm: ## Undeploy controller from the K8s cluster
 	$(KUSTOMIZE) build config/install | kubectl delete -f -
 
+OPERATOR_TYPE ?= standalone
```
iamniting, line 138: As we know this is a standalone operator only why can not we directly add the annotation in the CSV instead of building automation around it?

If we really need this, Lets add this variable in the `hack/make-bundle-vars.mk`

index.Dockerfile
```diff
@@ -5,6 +5,7 @@ ARG BUNDLE_IMGS=quay.io/ocs-dev/odf-operator-bundle:latest
 RUN opm index add \
 --bundles "${BUNDLE_IMGS}" \
 --out-dockerfile index.Dockerfile \
+--permissive \
```
iamniting, line 8: This can be squashed into other commit where this file was added

Makefile
```diff
@@ -43,27 +43,21 @@ fmt: ## Run go fmt against code.
...
+godeps-update: ## Run go mod tidy and go mod vendor.
 	go mod tidy && go mod vendor
 
-test: manifests generate fmt vet ## Run tests.
-	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo \
-		${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.8.3/hack/setup-envtest.sh
-	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); \
-		go test -coverprofile cover.out `go list ./... | grep -v "e2e"`
+test-setup: generate fmt vet godeps-update ## Run setup targets for tests
```
iamniting, line 52: We should also add `manifests` as it was present in the `test`

iamniting (changes requested):

index.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM quay.io/operator-framework/upstream-opm-builder:v1.19.5 AS builder
```
iamniting, line 1: Can you pls tell me why we need this, What is the motive behind adding the custom dockerfile instead of using the OPM binary directly? Also as we are not using OPM anymore we don't need opm target anymore. 

But I would still prefer using the opm binary instead of creating a dockerfile

iamniting (changes requested):

Makefile
jarrpa: This is specifically for the golang unit tests. Do we really need to generate YAML manifests for them? It's fine if yes, I'm just not aware of where we need that.

Makefile
jarrpa: I've been switching it to `non-standalone` for a few tests and experiments. Though I agree, this would be better in the `.mk` files.

index.Dockerfile
jarrpa: Valid.

index.Dockerfile
jarrpa: Specifically, this Dockerfile *already contains* the OPM binary. By using a container image, we completely avoid the problem of ensuring we have the correct version of the binary on the developer's system and all the directory path logic to keep it in the right place.

But you raise an interesting point. If we're not using OPM then what are we using to generate the CatalogSource index image? It's cool if it exists, I'm just not aware of any alternative.

Makefile
iamniting: `manifests` are mainly used in the bundle as far as know and we are calling it specifically in the bundle at many other places. So we should be good even if we do not add it.

index.Dockerfile
iamniting: Jose, I guess you misunderstood my point I mean to say is we should also remove a target in the make file which downloads the opm binary [1]. Also, we should stop calling the target in the catalog-build target [2].

[1]. https://github.com/red-hat-storage/odf-operator/blob/main/hack/make-tools.mk#L27-L30
[2]. https://github.com/red-hat-storage/odf-operator/blob/main/Makefile#L165

iamniting (approved):
/lgtm

iamniting:
looks like openshift-ci is not picking up the `${GINKGO}` value

Code:
```
cd e2e/odf && ${GINKGO} build && ./odf.test \
```

Error:
```
./hack/e2e-test.sh: line 7: build: command not found
make: *** [Makefile:63: e2e-test] Error 127
```

hack/e2e-test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+SCRIPT_DIR="$(dirname "$(realpath "$0")")"
```
iamniting, line 3: We are not using this var here.

hack/e2e-test.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -x
+
+cd e2e/odf && ${GINKGO} build && ./odf.test \
```
iamniting, line 7: let's look at the GINKGO why is it not picking up the value

iamniting (changes requested):

iamniting (approved):
/lgtm

PR #239, opened 2022-08-09
controllers: Do not verify IBM CSI driver CSV version

iamniting:
IBM CSI driver is backward compatiable with IBM odf-operator and should
be allowed to upgrade at any point of time. Which means we should not
check for specific version of its CSV.

https://bugzilla.redhat.com/show_bug.cgi?id=2104833


agarwal-mudit (approved):
/lgtm

PR #242, opened 2022-08-23
controllers: create cluster claims for DisasterRecovery

vbnrh:
This commit adds changes to create ACM resource "ClusterClaim" on the
cluster. This allows the hub cluster to read for data from its managed
clusters.

Signed-off-by: Vineet Badrinath <vineetbnath@gmail.com>


controllers/clusterclaims.go
```diff
@@ -0,0 +1,145 @@
...
+	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	clusterv1alpha1 "open-cluster-management.io/api/cluster/v1alpha1"
+)
+
+type ClusterClaimKey string
+
+const (
+	OdfVersion            ClusterClaimKey = "odf-version"
```
GowthamShanmugam, line 23: What do you think about adding odf as a prefix in front of all clusterClaim names? I feel it will avoid name conflict and safer to add these names in ACM Managed cluster

controllers/clusterclaims.go
```diff
@@ -0,0 +1,145 @@
...
+
+	creator := ClusterClaimCreator{
+		Logger:  logger,
+		Context: ctx,
+		Client:  r.Client,
+		Values:  make(map[ClusterClaimKey]string),
+	}
+
+	odfVersion, err := creator.getOdfVersion(instance.Namespace)
+	if len(odfVersion) == 0 || err != nil {
```
GowthamShanmugam, line 58: is there any particular reason for  "" to length conversion?

GowthamShanmugam:
LGTM

controllers/clusterclaims.go
vbnrh: It will help us identify that it belongs to odf but i don't think there will be name conflict for clusterclaim. 
For identification, i can add owner ref to the storagesystem and it will also help us with the uninstall. What do you think ?

controllers/clusterclaims.go
vbnrh: No reason, i just like to make sure that the string does not have any unwanted spaces.

controllers/clusterclaims.go
GowthamShanmugam: when you create cluster claim on odf cluster, it will create one status entry on ACM managed cluster CR, i think i will use the same which you used for cluster claim. so when it displayed on ACM  managed cluster CR, we need to makesure our ODF names dont have a conflict with others.

GowthamShanmugam: ya the owner ref is another good feature to cleanup

GowthamShanmugam (approved):

iamniting (commented):
Hello Vineet, This is out of scope for odf-operator. odf-operator is not aware of the ceph and all. Moreover, it also installs the IBM storagesystem. 

I think this should go to the ocs-operator.

GowthamShanmugam:
> Hello Vineet, This is out of scope for odf-operator. odf-operator is not aware of the ceph and all. Moreover, it also installs the IBM storagesystem.
> 
> I think this should go to the ocs-operator.

Here, We are syncing basic ODF information to the Hub cluster's ManagedCluster CR.
ManagedCluster CR is nothing but provides very high-level information about the managed cluster.
Once, the ODF cluster is managed by the ACM, this code will sync very the basic information about the ODF operator to the HUB cluster. So that the user can know about ODF from the Hub cluster itself. Also, DR UI can know about the managed cluster and its ODF details from ManagedCluster CR itself.

It can be any kind of ODF cluster, for now, we are syncing only OCS type StroageSystme info. In the future, we will sync IBM type also.

jarrpa:
...I don't see why this needs to be in the odf-operator. Looking at the [spec](https://github.com/open-cluster-management-io/api/blob/8d7cc32bed7b35d941e3ab65c186b27f6f70f8ba/cluster/v1alpha1/types.go#L114-L134), there's nothing to say it *needs* to be created by any particular operator/controller as long as it properly exists in the managed cluster. The only ClusterClaim that I see as valid for odf-operator would be the StorageSystem version. Everything else should be created by a similar controller in ocs-operator.

jarrpa (changes requested):
...I don't see why this needs to be in the odf-operator. Looking at the [spec](https://github.com/open-cluster-management-io/api/blob/8d7cc32bed7b35d941e3ab65c186b27f6f70f8ba/cluster/v1alpha1/types.go#L114-L134), there's nothing to say it *needs* to be created by any particular operator/controller as long as it properly exists in the managed cluster. The only ClusterClaim that I see as valid for odf-operator would be the StorageSystem claim. Everything else should be created by a similar controller in ocs-operator.

controllers/clusterclaims.go
```diff
@@ -0,0 +1,159 @@
...
+	StorageSystem *odfv1alpha1.StorageSystem
+}
+
+func (r *StorageSystemReconciler) ensureClusterClaims(instance *odfv1alpha1.StorageSystem, logger logr.Logger) error {
+	ctx := context.TODO()
+	crd := extensionsv1.CustomResourceDefinition{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: "clusterclaims.cluster.open-cluster-management.io"}, &crd)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("clusterclaims.cluster.open-cluster-management.io CRD not found. skipping creation of clusterclaims")
```
jarrpa, line 46: Remove this line, we don't need to spam the logs with this on every reconcile. We should only output when we're actually creating anything.

iamniting:
closing in favour of https://github.com/red-hat-storage/ocs-operator/pull/1789

PR #245, opened 2022-09-05
reconcile csiaddons subscription first

Rakshith-R:
CSIAddons has no dependencies and it provides
VR and VRC CRDs which will be required by
ocs-operator csv, hence reconcile first.

Signed-off-by: Rakshith R <rar@redhat.com>


iamniting:
We do not need this PR, Changing the order is not going to help us at all at this point. We will already be having the subscription installed via OLM as we also use OLM pkg dependencies.

PR #258, opened 2022-09-15
bundle: start creating file based catalog instead of sqlite

iamniting:
Sqlite-based catalogs and their related subcommands are deprecated.
Support for them will be removed in a future release.

Ref:
https://olm.operatorframework.io/docs/reference/file-based-catalogs/
https://olm.operatorframework.io/docs/reference/veneers/


iamniting:
The workflow is failing as code for accessing the secret is added into the same PR I guess.

jarrpa (commented):
Changes look good. Any reason to keep `index.Dockerfile` after this if all the checks are passing? You should also probably update (or just remove) the "comma-separated list" comment in the `.mk` file.

iamniting:
> Changes look good. Any reason to keep `index.Dockerfile` after this if all the checks are passing? You should also probably update (or just remove) the "comma-separated list" comment in the `.mk` file.

done, I missed it earlier

malayparida2000 (approved):
/lgtm

PR #261, opened 2022-09-19
render the bundle images outside of the container

iamniting:
github does not allow to use secrets in the pull requests. Which
prevent us to login to quay and push images to it. That can be pulled
inside the container while making a container image.

using opm outside of the container does allow to use the local bundle
images as well. Which help to pull odf bundle image from the local
registry.

Related PR: https://github.com/red-hat-storage/odf-operator/pull/258


malayparida2000 (approved):
/lgtm

PR #264, opened 2022-09-29
makefile: build on ARM64 machine

synarete:
Take GOOS and GOARCH make variables from local host (instead of hard-coded linux/amd64), in order to enable build on ARM based machines.

Tested on centos-stream9 aarch64 machine.

Signed-off-by: Shachar Sharon <ssharon@redhat.com>


iamniting (approved):
/lgtm

PR #277, opened 2022-12-05
Remove namespace dependency from the configmap and inject it in the deployment

iamniting:


malayparida2000 (approved):
/lgtm

PR #279, opened 2023-01-16
generate fusion data foundation operator bundle

umangachapagain:
Includes changes to the make targets to generate CSV for Fusion Data Foundation operator, build a bundle for this operator and to build a fusion operator catalog that includes this bundle.

It is generated based on existing ODF bundle resources and requires that they exist.

PR #281, opened 2023-01-31
bundle: update fusion bundle with approved content

umangachapagain:
Changes the name of fusion package to odf-operator. This is required for us to support seamless upgrades. OLM does not support upgrades between 2 different packages.

It also updates base CSV metadata with proper branding contents.

Enchances: #279


iamniting:
After deliberating among ourselves, we determined that because we are using the same name for the CSV, we do not need the entire set of changes. Will create a new PR with the fewest changes possible.

iamniting:
Here is a new PR I mentioned. https://github.com/red-hat-storage/odf-operator/pull/284

PR #283, opened 2023-02-02
revert the changes made for the fusion bundle

iamniting:
Previously, we attempted to use a separate CSV name for the fusion bundles, which necessitated these changes. Upgrades between different packages are not supported by OLM. So we don't need these changes anymore.

Will create a new PR with the minimal required changes.


umangachapagain (approved):
This reverts PR #279 and replaces #281 .

iamniting:
Here is a new PR I mentioned. https://github.com/red-hat-storage/odf-operator/pull/284

PR #287, opened 2023-02-07
console: configure securityContext for odf-console

SanjalKatiyar:
BZ:  https://bugzilla.redhat.com/show_bug.cgi?id=2166417
Issue: 115558: Misconfigured Security Context Attribute For A Pod Or Container

set securityContext for odf-console pod/container as per best practices.
Signed-off-by: SanjalKatiyar <sanjaldhir@gmail.com>

PR #289, opened 2023-02-08
Revert "console: configure securityContext for odf-console"

SanjalKatiyar:
This reverts commit 363419365ccd6c8fafe200e80b9d0bc0d9a232af.

SecurityContext: "readOnlyRootFilesystem: true" is causing odf-console pod crashloopbackoff. Nginx used downstream is trying to create some temp files but don't have access to do so, hence pod is never getting into "Running" status.

Will create a new PR without setting "readOnlyRootFilesystem" context for the container. And need to check why nginx wants such access.

Signed-off-by: SanjalKatiyar <sanjaldhir@gmail.com>


iamniting (approved):
It reverts https://github.com/red-hat-storage/odf-operator/pull/287

PR #290, opened 2023-02-08
Add/update securityContext for containers

SanjalKatiyar:
Signed-off-by: SanjalKatiyar <sanjaldhir@gmail.com>

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -473,6 +473,8 @@ spec:
                   capabilities:
                     drop:
                     - ALL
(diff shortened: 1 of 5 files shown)


bundle/manifests/odf-operator.clusterserviceversion.yaml
```diff
@@ -472,7 +472,7 @@ spec:
                   allowPrivilegeEscalation: false
                   capabilities:
                     drop:
-                    - all
```
iamniting, line 475: We do have other instances as well for this, lets update all of them

iamniting (changes requested):

iamniting (approved):

PR #301, opened 2023-03-16
controllers: do not use odf sub cache

iamniting:
Avoid using the cached sub of odf in a way that assumes it will always exist, because if someone deletes it, the odf operator will still have the same cache and will create new subs using the UID of the deleted odf sub. These new subs will then be marked for garbage collection by Kubernetes since the old odf sub UID no longer exists.

https://bugzilla.redhat.com/show_bug.cgi?id=2178619

### DIFF
--- a/controllers/defaults.go
+++ b/controllers/defaults.go
@@ -18,8 +18,6 @@ package controllers
 
 import (
 	"os"
-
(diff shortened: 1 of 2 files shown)


malayparida2000 (approved):
/lgtm

PR #307, opened 2023-04-05
console: securityContext: add 'readOnlyRootFilesystem'

alfonsomthd:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2166417

NGINX requires some dirs to be writable so we enable 2 `emptyDir` volumes for that purpose and update the NGINX config with the path of those dirs.

NOTE: when using the downstream image, we noticed this in the pod logs:
```/opt/app-root/etc/generate_container_user: line 6: /opt/app-root/etc/passwd: Read-only file system```
After a conversation with @SanjalKatiyar and Boris Ranto we decided to not take any action on this as it has no side effects on the behavior of the container.

### DIFF
--- a/console/nginx_conf.go
+++ b/console/nginx_conf.go
@@ -22,7 +22,7 @@ var NginxConf = `
 
 worker_processes auto;
 error_log /var/log/nginx/error.log;
-pid /run/nginx.pid;
+pid /var/lib/nginx/tmp/nginx.pid;
 
 # Load dynamic modules. See /usr/share/doc/nginx/README.dynamic.
 include /usr/share/nginx/modules/*.conf;
--- a/console/nginx_conf.go
+++ b/console/nginx_conf.go
@@ -32,6 +32,13 @@ events {
 }
 
 http {
+    # Use directories writable by unprivileged users.
+    client_body_temp_path /var/lib/nginx/tmp/client_temp;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: console/nginx_conf.go [hunk 1]
```diff
@@ -22,7 +22,7 @@ var NginxConf = `
 
 worker_processes auto;
 error_log /var/log/nginx/error.log;
-pid /run/nginx.pid;
+pid /var/lib/nginx/tmp/nginx.pid;
```
SanjalKatiyar, line 25:
```suggestion
pid /var/lib/nginx/run/nginx.pid;
```
can't we keep naming same as original ?

SanjalKatiyar, line 25:
```suggestion
pid /var/log/nginx/nginx.pid;
```
what if we add it to the same location as error.log ?
that way we don;t have to provide write permission for 2 separate paths...

SanjalKatiyar: okay... if it needs access to this path anyway then we are good...

SanjalKatiyar:
/retitle [WIP] console: securityContext: add 'readOnlyRootFilesystem'

SanjalKatiyar:
/lgtm
cc @iamniting (no need to backport)

iamniting (approved):

PR #310, opened 2023-04-17
Adds container liveness checks for odf-console

bipuladh:
Screenshot from 2023-04-17 14-23-53

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -460,6 +460,13 @@ spec:
             spec:
               containers:
               - image: quay.io/ocs-dev/odf-console:latest
+                livenessProbe:
+                  httpGet:
+                    path: /plugin-manifest.json
+                    port: 9001
+                    scheme: HTTPS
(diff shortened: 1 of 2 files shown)


iamniting (approved):

iamniting:
/hold

By the way, are you sure this file will exist in the upstream and downstream containers both?

bipuladh:
It is important to note that the odf-console project produces identical files for both upstream and downstream. While we employ a node-based image for upstream and an nginx image for internal builds, this does not impact the contents of the files. As both of these platforms function as web servers, the GET requests for these probes perform well with either image. Regardless of the build environment, the files that make up any given image remain unaltered. It is worth mentioning that the contents of files may vary for MCO; however, the /plugin-manifest.json file remains consistent even for MCO. In addition, all dynamic plugins utilize the ConsoleRemotePlugin webpack plugin, which generates the same filename for all plugins, namely plugin-manifest.json. This file name is hardcoded, as can be verified by examining https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/src/constants.ts#L2

PR #316, opened 2023-06-07
makefile: block OCP upgrades beyond the X+1 version

iamniting:
The annotation will allow OCP upgrade till given max OCP version which
is X+1. For example the ODF version is 4.13 the max OCP version would
be 4.14.
    
Ref: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html-single/operators/index#:~:text=If%20you%20know,the%20cluster%20version

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -37,6 +37,7 @@ metadata:
     containerImage: quay.io/ocs-dev/odf-operator:latest
     description: OpenShift Data Foundation provides a common control plane for storage
       solutions on OpenShift Container Platform.
+    olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14"}]'
     olm.skipRange: ""
     operatorframework.io/initialization-resource: |-
       {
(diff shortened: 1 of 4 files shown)


bundle/manifests/odf-operator.clusterserviceversion.yaml
```diff
@@ -37,6 +37,7 @@ metadata:
     containerImage: quay.io/ocs-dev/odf-operator:latest
     description: OpenShift Data Foundation provides a common control plane for storage
       solutions on OpenShift Container Platform.
+    olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14"}]'
```
umangachapagain, line 40:
```suggestion
    olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14.0"}]'
```

umangachapagain: Is this how the spec is ? Seems weird.

iamniting: Yes it is, Pls refer to the link in the description box. Also it should be `4.14` not `4.14.0`

umangachapagain (approved):

PR #322, opened 2023-06-09
console: Fixes reconciliation logic for console ConfigMap

bipuladh:
The mutateFunciton is updated to modify the configmap

Currently, when ODF is upgraded the ConfigMap data does not get updated. Hence the container crashes. This updates the configmap which should stop the crash. 
The configmap holds configuration related to Nginx. 
In ODF 4.13 the pid directive was pointing to `/run/nginx.pid` which was updated to `/var/lib/nginx/tmp/nginx.pid`.
Based on above change we set `readOnlyRootFilesystem` to `true`. 
As the ConfigMap was not updated Nginx tried creating a file at `/run/nginx.pid`.

### DIFF
--- a/controllers/clusterversion_controller.go
+++ b/controllers/clusterversion_controller.go
@@ -88,6 +88,7 @@ func (r *ClusterVersionReconciler) ensureConsolePlugin(clusterVersion string) er
 	logger := log.FromContext(context.TODO())
 	// The base path to where the request are sent
 	basePath := console.GetBasePath(clusterVersion)
+	nginxConf := console.GetNginxConfiguration()
 
 	// Get ODF console Deployment
 	odfConsoleDeployment := console.GetDeployment(OperatorNamespace)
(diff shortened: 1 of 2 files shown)


controllers/clusterversion_controller.go
```diff
@@ -102,6 +103,10 @@ func (r *ClusterVersionReconciler) ensureConsolePlugin(clusterVersion string) er
 	// Create/Update ODF console ConfigMap (nginx configuration)
 	odfConsoleConfigMap := console.GetNginxConfConfigMap(OperatorNamespace)
 	_, err = controllerutil.CreateOrUpdate(context.TODO(), r.Client, odfConsoleConfigMap, func() error {
+		if odfConsoleConfigMapData := odfConsoleConfigMap.Data["nginx.conf"]; odfConsoleConfigMapData != nginxConf {
```
iamniting, line 106: Instead of an if why can not we always assign the value. It won't have any effect, right?

iamniting (changes requested):

iamniting (approved):

controllers/clusterversion_controller.go
iamniting: Resolving it, we sorted it over chat

PR #325, opened 2023-07-05
Revert "makefile: add fusion bases required to generate fusion bundle"

iamniting:
This reverts commit c9232375d5f57f67b13d3bb7272c46b7ff3ed10c.


config/manifests/kustomization.yaml
```diff
@@ -1,8 +1,11 @@
 # These resources constitute the fully configured set of manifests
 # used to generate the 'manifests/' directory in a bundle.
-
```
malayparida2000, line 3: Why are we removing the apiVersion & Kind, Aren't they necessary?

iamniting: no, they are not. This is how the scaffolding was created earlier. As it is revert git is reverting everything. They get populated if and only if we change the file. Which was the case when we added the fusion.

malayparida2000: ack

malayparida2000 (approved):
/lgtm

PR #344, opened 2023-11-20
add annotation for STS support

Nikhil-Ladha:
Added an annotation to odf-operator CSV for STS support.

Ref: https://issues.redhat.com/browse/RHSTOR-5097


Nikhil-Ladha:
@iamniting updated the PR to only add a new annotation to the CSV

iamniting (approved):

PR #349, opened 2023-12-06
set ROLEARN env variable in noobaa subscription

Nikhil-Ladha:
If ROLEARN env variable is set, then pass on that env variable to noobaa subscription.

Ref: https://issues.redhat.com/browse/RHSTOR-5139


iamniting (changes requested):
Can you pls change `CheckExistingSubscriptions` to always update the env variable.

PR #355, opened 2023-12-08
Surface dependent Operator upgradeable conditions

leelavg:
- odf-operator brings dependencies using OLM
- this PR ensures odf-operator sets it's own operator condition based on upgradeability of it's dependents

[RHSTOR-5073]

PR #359, opened 2023-12-20
Bug 2255315:[release-4.14] Surface dependent Operator upgradeable conditions

leelavg:
manual cherry-pick of #355, the difference is updating go.mod with only required modules


leelavg:
taken in #369

PR #368, opened 2024-02-05
controllers: replicate odf tolerations to child subscriptions

iamniting:

### DIFF
--- a/controllers/subscriptions.go
+++ b/controllers/subscriptions.go
@@ -42,16 +42,27 @@ import (
 // because fake.Client does not support them.
 func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operatorv1alpha1.Subscription) (*operatorv1alpha1.Subscription, error) {
 
-	subsList := &operatorv1alpha1.SubscriptionList{}
-	err := cli.List(context.TODO(), subsList, &client.ListOptions{Namespace: desiredSubscription.Namespace})
+	odfSub, err := GetOdfSubscription(cli)
 	if err != nil {
 		return nil, err
 	}
 
+	if odfSub.Spec.Config == nil {
+		odfSub.Spec.Config = &operatorv1alpha1.SubscriptionConfig{}
+	}
+
+	subsList := &operatorv1alpha1.SubscriptionList{}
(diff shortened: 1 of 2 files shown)


controllers/subscriptions.go
```diff
@@ -42,39 +42,99 @@ import (
...
 			if actualSub != nil {
 				foundSubs := []string{actualSub.Name, sub.Name}
 				return nil, fmt.Errorf("multiple Subscriptions found for package '%s': %v", pkgName, foundSubs)
 			}
 			actualSub = &subsList.Items[i]
 			actualSub.Spec.Channel = desiredSubscription.Spec.Channel
 
-			// If the config is not set, only then set it to the desired value => allow user to override
-			if actualSub.Spec.Config == nil {
+			if actualSub.Spec.Config == nil && desiredSubscription.Spec.Config == nil {
```
umangachapagain, line 73: Why not just update https://github.com/red-hat-storage/odf-operator/blob/1969776bdf4ad5bc194cdca0a953fa473e3b6d82/controllers/subscriptions.go#L260 to inject tolerations instead of all these checks?

Add tolerations to desired subscription, and during reconcile merge actual and desired subs. Wouldn't that be simpler?

umangachapagain (changes requested):

controllers/subscriptions.go
```diff
@@ -42,39 +42,99 @@ import (
...
 func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operatorv1alpha1.Subscription) (*operatorv1alpha1.Subscription, error) {
 
+	odfSub, err := GetOdfSubscription(cli)
+	if err != nil {
+		return nil, err
+	}
+
+	if odfSub.Spec.Config == nil {
+		odfSub.Spec.Config = &operatorv1alpha1.SubscriptionConfig{}
+	}
```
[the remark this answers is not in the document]
iamniting: It is not gonna update. It is just to save from the panic in the below code.

controllers/subscriptions.go
iamniting: Actually that will require extensive changes in the code, we will have to change the signature of multiple funcs.

umangachapagain (approved):

PR #370, opened 2024-02-09
bundle: Modify the name and reference of the sample storagesystem

iamniting:
In certain instances customers unintentionally create two storagesystems to the same storagecluster. This becomes problematic when attempting to delete one storagesystem, as it triggers the deletion of the associated storagecluster. By changing the name and reference, we aim to prevent this issue from occurring.

It will make sure that even if they create the storagesystem from the OCP console it wont refer to the same storagecluster.

Bugs:
https://bugzilla.redhat.com/show_bug.cgi?id=2263441
https://bugzilla.redhat.com/show_bug.cgi?id=2254343

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -8,12 +8,12 @@ metadata:
           "apiVersion": "odf.openshift.io/v1alpha1",
           "kind": "StorageSystem",
           "metadata": {
-            "name": "ibm-flashsystemcluster-storagesystem",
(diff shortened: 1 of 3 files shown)


umangachapagain (commented):
This is avoiding the problem and not actually fixing it.
Nothing stops me from creating 2 StorageSystems pointing to the same StorageCluster.

I see 2 options for a complete fix:
1. Admission webhook to reject such requests immediately.
2. Validate that StorageCluster does not already have a different owner before taking it's ownership. Current behavior of owning the StorageCluster without validation is a bug.

umangachapagain (approved):
We will add more fixes as a follow-up. This is a good start.

/lgtm

PR #374, opened 2024-02-09
Bug 2263475:[release-4.12] bundle: Modify the name and reference of the sample storagesystem

iamniting:
This is an manual cherry-pick of #370


Nikhil-Ladha (approved):
/lgtm

PR #375, opened 2024-02-09
Bug 2263476:[release-4.11] bundle: Modify the name and reference of the sample storagesystem

iamniting:
This is an manual cherry-pick of #370


iamniting:
4.11 is EOL

PR #383, opened 2024-03-22
bundle: do not reconcile ocs-client subscription

leelavg:
ocs-client operator handles it's own subscription and odf-operator trying to reconcile ocs-client subscription can cross paths and get into unforseen deadlock.


iamniting:
Will be done with the #393

PR #384, opened 2024-03-25
Update prometheus operator from v4.10.0 to v4.15.0

ezio-auditore:


iamniting:
This is being done with #392

PR #385, opened 2024-03-26
controller: reuse the context from reconcile

Madhu-1:
We dont need to create a new context for each time for a reconcile we should use the ctx from the Reconcile for all the dependent operations.

### DIFF
--- a/controllers/clusterversion_controller.go
+++ b/controllers/clusterversion_controller.go
@@ -37,6 +37,7 @@ import (
 
 // ClusterVersionReconciler reconciles a ClusterVersion object
 type ClusterVersionReconciler struct {
+	ctx context.Context
 	client.Client
 	Scheme      *runtime.Scheme
 	ConsolePort int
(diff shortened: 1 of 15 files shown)


Madhu-1:
@iamniting where i can find details about the test failure?

iamniting (changes requested):
Can we save the context to the recociler func and use it from there. I will save us from passing ctx here and there if the func is on top of reconciler struct.

Madhu-1:
> Can we save the context to the recociler func and use it from there. I will save us from passing ctx here and there if the func is on top of reconciler struct.

sorry didn't get the suggestion, Do you mean saving as a global variable? if yes, i would not do it as its against the practice. main goal is that we should not create context when reconciling instead we should use the context created for the `Reconcile`

Madhu-1:
> Can we save the context to the recociler func and use it from there. I will save us from passing ctx here and there if the func is on top of reconciler struct.

got it now, let me take that in to consideration and do it

PR #387, opened 2024-03-27
console: adds proxy declaration for ROSA monitoring stack

bipuladh:


console/console.go
```diff
@@ -105,6 +105,16 @@ func GetConsolePluginCR(consolePort int, serviceNamespace string) *consolev1alph
...
+				{
+					Type:  consolev1alpha1.ProxyTypeService,
+					Alias: "rosa-prometheus",
+					Service: consolev1alpha1.ConsolePluginProxyServiceConfig{
+						Name:      "prometheus",
+						Namespace: serviceNamespace,
+						Port:      9339,
+					},
+					Authorize: false,
+				},
```
SanjalKatiyar, line 117: can we maybe add it optionally (only if it is ROSA infra) ?

bipuladh: that requires us to detect platform and would add unnecessary complications to our code.

iamniting (approved):

PR #388, opened 2024-03-28
makefile: update odf go mod depndancies & odf version to 4.16 .0

malayparida2000:
Also updated ocs version being used to 4.16

PR #391, opened 2024-03-29
makefile: Introduce Separate Variable for CSV Version

iamniting:
sometimes image tags are not same as the csv version. which makes it impossible to use the same variable for the image tag and csv version.

should be merged after #388


malayparida2000 (approved):
/lgtm

PR #393, opened 2024-04-02
controllers: Do not update client operator in provider mode

iamniting:
Client operator manages its own upgrade life cycle in the provider mode, Do not update its channel in the provider mode.

### DIFF
--- a/controllers/storagesystem_controller.go
+++ b/controllers/storagesystem_controller.go
@@ -34,6 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v4/v1"
 	odfv1alpha1 "github.com/red-hat-storage/odf-operator/api/v1alpha1"
 	"github.com/red-hat-storage/odf-operator/metrics"
 	"github.com/red-hat-storage/odf-operator/pkg/util"
@@ -219,7 +220,10 @@ func (r *StorageSystemReconciler) ensureSubscriptions(instance *odfv1alpha1.Stor
 
 func (r *StorageSystemReconciler) isVendorCsvReady(instance *odfv1alpha1.StorageSystem, logger logr.Logger) error {
 
-	csvNames := GetVendorCsvNames(instance.Spec.Kind)
+	csvNames, err := GetVendorCsvNames(r.Client, instance.Spec.Kind)
+	if err != nil {
+		return err
+	}
 
 	var returnErr error
 	for _, csvName := range csvNames {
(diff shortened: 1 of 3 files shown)


controllers/subscriptions.go
```diff
@@ -102,6 +114,23 @@ func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operator
 	return desiredSubscription, nil
 }
 
+func isProviderServer(cli client.Client) (bool, error) {
```
leelavg, line 117: not sure `isProviderServer` sounds correct, to be abstract can use `isProviderMode` or to be exact can use `isAllowRemoteStorageConsumersSet`

controllers/subscriptions.go
```diff
@@ -102,6 +114,23 @@ func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operator
...
+		return false, err
+	}
+
+	for _, storagecluster := range storageclusters.Items {
+		if storagecluster.Spec.AllowRemoteStorageConsumers {
+			return true, nil
+		}
+	}
+
+	return false, nil
```
leelavg, line 131: sry, this seems a lot of work to find the fields values, not mandatory and here's what I'm thinking, `cli.List` will try to list storageclusters in all namespaces and `storagecluster` in for loop is making a copy of every storagecluster in the Items.

It might be good to limit the result to operator ns and taken an index into Items? again, only a suggestion.

controllers/storagesystem_controller.go
```diff
@@ -259,6 +263,7 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&odfv1alpha1.StorageSystem{}, builder.WithPredicates(generationChangedPredicate)).
 		Owns(&operatorv1alpha1.Subscription{}, builder.WithPredicates(generationChangedPredicate, ignoreCreatePredicate)).
+		Owns(&ocsv1.StorageCluster{}, builder.WithPredicates(generationChangedPredicate)).
```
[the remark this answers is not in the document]
iamniting: Yes

controllers/subscriptions.go
```diff
@@ -51,6 +52,14 @@ func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operator
 		odfSub.Spec.Config = &operatorv1alpha1.SubscriptionConfig{}
 	}
 
+	var isProvider bool
+	if desiredSubscription.Spec.Package == OcsClientSubscriptionPackage {
```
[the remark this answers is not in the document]
iamniting: Yes it is defined in the defaults.go

controllers/subscriptions.go
```diff
@@ -68,7 +77,10 @@ func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operator
 				return nil, fmt.Errorf("multiple Subscriptions found for package '%s': %v", pkgName, foundSubs)
 			}
 			actualSub = &subsList.Items[i]
-			actualSub.Spec.Channel = desiredSubscription.Spec.Channel
+
+			if !isProvider {
+				actualSub.Spec.Channel = desiredSubscription.Spec.Channel
+			}
```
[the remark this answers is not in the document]
iamniting: no, it does not overwrite metadata.

controllers/subscriptions.go
iamniting: I will change it to `isProviderMode` as we always refer it as provider.

controllers/subscriptions.go
iamniting: I would say let it be. We may extend the func later on.

leelavg (approved):

PR #396, opened 2024-04-05
bundle: add readOnlyRootFilesystem for odf-operator

iamniting:

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2273386

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -35,7 +35,7 @@ metadata:
     categories: Storage
     console.openshift.io/plugins: '["odf-console"]'
     containerImage: quay.io/ocs-dev/odf-operator:latest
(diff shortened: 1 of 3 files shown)


SanjalKatiyar:
LGTM.

But, I remember when we simply added `readOnlyRootFilesystem: true` for `odf-console` pod, we had to revert the changes temporarily, so better to test this once (if not already) on close to prod-like environment (rhceph-dev) and verify if all pods (manager/proxy/console etc) are running as expected.

iamniting:
> LGTM.
> 
> But, I remember when we simply added `readOnlyRootFilesystem: true` for `odf-console` pod, we had to revert the changes temporarily (had to do much more changes to make it work), so better to test this once (if not already) on close to prod-like environment (rhceph-dev) and verify if all pods (manager/proxy/console etc) are running as expected.

Tested, It is working well.

Nikhil-Ladha (approved):
lgtm

PR #394, opened 2024-04-02
create a mechanism to deploy odf using operator-sdk

iamniting:
Signed-off-by: Nitin Goyal <nigoyal@redhat.com>

PR #401, opened 2024-04-24
create a mechanism to deploy odf using operator-sdk

iamniting:
This is an manual cherry-pick of #394


iamniting:
It is purely tests related PR, It does not have any effect on the actual code. Will merge it without the BZ.

PR #402, opened 2024-04-24
Bug 2276913:[release-4.16] controllers: do not skip rook subscription reconcile

iamniting:
skipping rook subscription reconcile lead to the tolerations being absent from the rook subscription.

This fix is only required in 4.16 due to the #399


Nikhil-Ladha (approved):
/lgtm

PR #417, opened 2024-05-21
Add "ConsoleCLIDownload" to the bundle

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-5204

PR #418, opened 2024-05-21
Add the option to download the ODF CLI tool from the UI

SanjalKatiyar:
https://issues.redhat.com/browse/RHSTOR-5204

No straight-forward/correct way of CLI Tool CR deletion (clean-up) is currently possible (**during operator uninstallation**).
1. Tried deleting from `clusterversion_controller`, as part of reconcile (https://github.com/red-hat-storage/odf-operator/pull/410#discussion_r1607017878).
2. Tried adding CR as part of the ODF bundle for OLM to maintain (https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/bundle/supported_resources.go#L21 & https://github.com/red-hat-storage/odf-operator/pull/417#discussion_r1608233489).
3. One other way could be to add a finaliser to the manager pod which will be removed by `clusterversion_controller` once clean-up is done, but dropped this approach entirely after discussion due to hack-ish nature.

This PR handles CLI Tool CR creation as part of `clusterversion_controller` and deletion will be a manual step which needs to be documented.

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -35,7 +35,7 @@ metadata:
     categories: Storage
     console.openshift.io/plugins: '["odf-console"]'
     containerImage: quay.io/ocs-dev/odf-operator:latest
-    createdAt: "2024-05-15T04:40:51Z"
+    createdAt: "2024-05-21T10:23:39Z"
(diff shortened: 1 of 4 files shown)


SanjalKatiyar:
cc @umangachapagain @iamniting @bipuladh (PTAL)

iamniting (approved):

PR #422, opened 2024-06-06
Bug 2296533:[release-4.14] api: update ocs-operator pkg version

Nikhil-Ladha:
Update ocs-operator pkg version to include new api changes
Ref BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2276694


Nikhil-Ladha:
> Can you move the new require block in the existing require block?
> 
> Can we reduce these other packages update?

I have moved the require modules to one section, but other package updates can't be controlled they come automatically due to ocs-operator upgrade as dependencies.

iamniting (approved):

iamniting (changes requested):
Is it possible to reduce these changes and only updates OCS?

Nikhil-Ladha:
> Is it possible to reduce these changes and only updates OCS?

Only ocs-operator is updated, rest of the changes are brought along by ocs-op itself.

iamniting (approved):

PR #424, opened 2024-06-06
Bug 2296521: [release-4.16] api: update ocs-operator pkg version

Nikhil-Ladha:
Update ocs-operator pkg version to include new api changes
Ref BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2276694

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/openshift/custom-resource-status v1.1.2
 	github.com/operator-framework/api v0.22.0
 	github.com/prometheus/client_golang v1.19.0
-	github.com/red-hat-storage/ocs-operator/api/v4 v4.0.0-20240328064129-39415c481f57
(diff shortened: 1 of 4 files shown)


iamniting (approved):

Nikhil-Ladha:
@iamniting I guess we can merge this (and other backports) PR without a bz, right?

iamniting:
> @iamniting I guess we can merge this (and other backports) PR without a bz, right?

no we need BZ

iamniting:
> @iamniting we are using `gingko@latest` in `make-tools` i.e, causing to download the latest gingko version which requires golang 1.22. I think we should pin to a version similar to one we use in go.mod. WDYT? Ref: https://github.com/red-hat-storage/odf-operator/blob/release-4.16/hack/make-tools.mk#L17

Yes, we should do that, Can you pls create the required PRs in all branches I will merge them without the bugs.

Nikhil-Ladha:
Here's the PR: #484

PR #425, opened 2024-06-06
Update ocs op pkg and bundle image

Nikhil-Ladha:
Update ocs-operator pkg and bundle image to include latest api changes


go.mod
```diff
@@ -22,22 +22,26 @@ require (
 	sigs.k8s.io/controller-runtime v0.17.2
 )
 
+require (
+	github.com/Masterminds/semver/v3 v3.2.1 // indirect
+	github.com/go-jose/go-jose/v4 v4.0.1 // indirect
+)
```
iamniting, line 28: pls move this require block in the existing require block

iamniting (changes requested):

go.mod
Nikhil-Ladha: Done

iamniting (approved):

PR #440, opened 2024-07-12
[release-4.15] create a mechanism to deploy odf using operator-sdk

iamniting:
This is an manual cherry-pick of #394 #412


hack/install-odf.sh
```diff
@@ -0,0 +1,31 @@
...
+
+oc wait --timeout=5m --for jsonpath='{.status.phase}'=Succeeded -n "$INSTALL_NAMESPACE" csv $CSV_NAMES
+
+oc wait --timeout=5m --for condition=Available -n "$INSTALL_NAMESPACE" deployment \
+    csi-addons-controller-manager \
+    noobaa-operator \
+    ocs-operator \
+    odf-console \
+    odf-operator-controller-manager \
+    rook-ceph-operator
```
Nikhil-Ladha, line 31: We have the `ux-backend-server` deployment in 4.15 as well.

.github/workflows/docker-build.yaml
```diff
@@ -72,3 +72,5 @@ jobs:
       run: |
         docker run -d -p 5000:5000 --name registry registry:2
         make bundle-build bundle-push catalog-build
```
Nikhil-Ladha, line 74: We also need to update add the `catalog-deps-build` command here, right?

malayparida2000 (approved):
/lgtm

.github/workflows/docker-build.yaml
Nikhil-Ladha: Also, to the `docker-push` action

iamniting: No there is no need for that, That is required only for the openshift CI

Nikhil-Ladha: Ok, I see we have that in 4.16 and main branch. So, shall we remove that from there as well?

iamniting: It's okay we can keep it there, It is not required at least on the old branch.

Nikhil-Ladha (approved):
/lgtm

.github/workflows/docker-build.yaml
iamniting: I have backported that commit as well.

Nikhil-Ladha: Cool 😅

Nikhil-Ladha (commented):
/lgtm

malayparida2000 (approved):

PR #450, opened 2024-07-19
controllers: Shut down OCS client CSV pods if not in provider mode

iamniting:
The ocs-client-operator is a dependency of the odf-operator, which gets installed regardless of whether the cluster is in provider mode. However, the client operator is not needed in non-provider mode. This change ensures that the client operator is shut down if the cluster is not in provider mode.

https://issues.redhat.com/browse/DFBUGS-17
https://bugzilla.redhat.com/show_bug.cgi?id=2299443

### DIFF
--- a/controllers/storagesystem_controller.go
+++ b/controllers/storagesystem_controller.go
@@ -263,7 +263,9 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&odfv1alpha1.StorageSystem{}, builder.WithPredicates(generationChangedPredicate)).
 		Owns(&operatorv1alpha1.Subscription{}, builder.WithPredicates(generationChangedPredicate, ignoreCreatePredicate)).
-		Owns(&ocsv1.StorageCluster{}, builder.WithPredicates(generationChangedPredicate)).
+		// Although we own the storage cluster, we are not a controller owner.
+		// Not being a controller owner requires us to pass builder.MatchEveryOwner.
+		Owns(&ocsv1.StorageCluster{}, builder.MatchEveryOwner, builder.WithPredicates(generationChangedPredicate)).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
 		Complete(r)
 }
--- a/controllers/subscriptions.go
+++ b/controllers/subscriptions.go
@@ -20,11 +20,15 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 
 	"go.uber.org/multierr"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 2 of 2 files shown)


controllers/subscriptions.go
```diff
@@ -259,9 +261,48 @@ func GetVendorCsvNames(cli client.Client, kind odfv1alpha1.StorageKind) ([]strin
...
+		// This can result in these operators lagging behind the odf-operator, with different CSV versions.
+		// Therefore, we need to fetch the CSV name from the operator deployment.
+		// We are only fetching the ocs-client-operator CSV name and ignoring the csiaddons
+		// Because the ocs-client-operator is essential when enabling provider mode as we need to bring it up.
+
+		// Fetch the CSV name from the client operator deployment and append it to the csv list
+		if isProvider {
+			// get ocs-client-operator deployment with label
+			deployments := &appsv1.DeploymentList{}
+			err = cli.List(context.TODO(), deployments, &client.ListOptions{
```
nb-ohad, line 283: Preferably let the caller pass a context and use it.
It does effect, cancelability, logging and stack traces

controllers/subscriptions.go
```diff
@@ -259,9 +261,48 @@ func GetVendorCsvNames(cli client.Client, kind odfv1alpha1.StorageKind) ([]strin
...
+				Namespace:     OperatorNamespace,
+				LabelSelector: labels.SelectorFromSet(map[string]string{"app": "ocs-client-operator"}),
+			})
+
+			if err != nil {
+				return csvNames, err
+			}
+
+			if len(deployments.Items) == 0 {
+				return csvNames, fmt.Errorf("ocs-client-operator deployment not found")
```
nb-ohad, line 293: The deployment name/label `ocs-client-operator` is repeating multiple times in the code (as a label name and in the error messages), please consider moving it into a const.

controllers/subscriptions.go
```diff
@@ -259,9 +261,48 @@ func GetVendorCsvNames(cli client.Client, kind odfv1alpha1.StorageKind) ([]strin
...
+			}
+
+			// get owner ref from deployment
+			ownerRef := deployments.Items[0].OwnerReferences
+			if len(ownerRef) == 0 {
+				return csvNames, fmt.Errorf("ocs-client-operator deployment owner reference not found")
+			}
+
+			// get csv name from owner ref
+			csvNames = append(csvNames, ownerRef[0].Name)
```
nb-ohad, line 303: Don't assume a single owner ref, search for it using `slices.Index` or `slices.IndexFunc`

controllers/subscriptions.go
```diff
@@ -285,6 +326,24 @@ func EnsureVendorCsv(cli client.Client, csvName string) (*operatorv1alpha1.Clust
...
+			if err != nil {
+				return err
+			}
+
+			var replicas int32 = 0
+			if isProvider {
+				replicas = 1
+			}
+
+			for i := range csvObj.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
```
nb-ohad, line 342: Are we sure nothing in this long path is a pointer and has the possibility to be `nil`?
It any of it is a pointer, then consider having a nil check before accessing to make sure we do not crush the pod in runtime for some edge case.

nb-ohad (changes requested):

controllers/subscriptions.go
iamniting: Actually we need a lot of changes for that in multiple places, Lets leave that for later.

controllers/subscriptions.go
iamniting: There is no pointer.

controllers/storagesystem_controller.go
```diff
@@ -263,7 +265,30 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+									Namespace: sc.Namespace,
+								}},
+							}
+						}
+					}
+
+					return []reconcile.Request{}
+				},
+			),
+			builder.WithPredicates(generationChangedPredicate)).
```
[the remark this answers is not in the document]
iamniting: Done

controllers/subscriptions.go
iamniting: Done

controllers/subscriptions.go
iamniting: It is used once in the label and twice in the error. As the string is used only once in the label lets keep it as it is. I want to ignore variables in errors it just reduces the readability.

controllers/subscriptions.go
nb-ohad: Ok

controllers/subscriptions.go
```diff
@@ -259,9 +262,52 @@ func GetVendorCsvNames(cli client.Client, kind odfv1alpha1.StorageKind) ([]strin
...
+			})
+			if ownerRefIndex == -1 {
+				return csvNames, fmt.Errorf("ClusterServiceVersion owner reference not found in ocs-client-operator deployment")
+			}
+
+			ownerRef := deployments.Items[0].OwnerReferences[ownerRefIndex]
+
+			// get csv name from owner ref
+			csvNames = append(csvNames, ownerRef.Name)
+			OcsClientSubscriptionStartingCSV = ownerRef.Name
```
nb-ohad, line 309: You are inconsistent with the naming convention. Both OCS and CSV are abbreviations but, you use camel-case for OCS and use uppercase for CSV. In most new operators we use camel-case for everything to prevent the issue of having two or more abbreviations one after the other, which in the case of the upper case makes the variable/field/func name unreadable.  

Regardless of which letter-casing you decide to use, the request is to stay consistent

nb-ohad, line 309: The CSV does not have to be the starting CSV, can we change the name to `OcsClientOperatorOwnerCsv`?

controllers/subscriptions.go
nb-ohad: Ok, but please consider the following:
- This is a prime candidate for a future bug where the selected label changes but the error message is stale. 
- I don't think that a format for a message makes it more difficulty from a readability standpoint to the point we must avoid it. As a matter of fact, it is a common practice all over any golang repo I ever worked on, so we know it does not create a readability issue

nb-ohad (changes requested):

controllers/subscriptions.go
iamniting: `OcsClientSubscriptionStartingCSV` is a global variable which is already defined in the package.

nb-ohad: So what if it is already defined, you are now rewriting the entire code that uses it. A variable with a non-fitting name confuses the reader.
Calling this a starting CSV is just wrong, and for someone coming from the outside you are planting the wrong assumptions in their mind

PR #462, opened 2024-08-12
controllers: delete the webhook created by the ocs-client-operator

iamniting:
Delete the webhook created by the ocs-client-operator as the client operator is already down in the 4.16.1 and it can not delete the webhook itself. Deleting the webhook will allow us the upgrades from 4.16 to 4.17. Otherwise it will always hit an error "failed calling webhook" while updating the channel of the ocs-client-operator subscription.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2304073

### DIFF
--- a/bundle/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/odf-operator.clusterserviceversion.yaml
@@ -159,6 +159,12 @@ spec:
           - update
           - watch
         - apiGroups:
+          - admissionregistration.k8s.io
+          resources:
+          - validatingwebhookconfigurations
+          verbs:
+          - delete
+        - apiGroups:
           - apiextensions.k8s.io
           resources:
           - customresourcedefinitions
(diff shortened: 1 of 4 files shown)


iamniting:
@leelavg I am deleting the webhook where I am verifying the CSV and making the replica count of the OCS-client-operator to 0. This step is after we update the subscription. 

Considering that the fix will go in the 4.16 it will work fine as we won't require a subscription change from 4.16 to 4.16 upgrade.

iamniting:
> > @leelavg I am deleting the webhook where I am verifying the CSV and making the replica count of the OCS-client-operator to 0. This step is after we update the subscription.
> > Considering that the fix will go in the 4.16 it will work fine as we won't require a subscription change from 4.16 to 4.16 upgrade.
> 
> * Ack, until we get the 4.16 build w/ this fix, change in 4.17 isn't effective. Understood correctly?

Yes correct.

leelavg (approved):
/lgtm

PR #474, opened 2024-09-12
controllers: scale down ocs-client-op csv in non-provider mode

leelavg:
t0: odf-op w/ X.Y.Z is installed and all dependecies at same version will be installed
t1: odf-op is upgraded from X.Y.Z to X.(Y+1).0 and update will happen for odf-op unless there are explicit not upgradeable conditions set
t2: dependencies will first get updated from X.Y.Z to X.Y.(Z+1) if the channel has the update, now if one of the dependencies is stuck odf-op keeps trying to check for upgraded install plan which will not be created due to previous install plan for dependencies not being fulfilled.

now, odf-op always tries to scale down client-op if not in provider mode in above scenario t1 & t2 aren't strictly linear and odf-op ended up scaling down X.Y.Z version of client-op but not X.Y.(Z+1) which will never reach running state if cluster is configured as non-provider mode.

current PR makes sure that client-op csv & deployments are always scaled down for all non-provider cases.

### DIFF
--- a/controllers/storagesystem_controller.go
+++ b/controllers/storagesystem_controller.go
@@ -220,6 +220,9 @@ func (r *StorageSystemReconciler) ensureSubscriptions(instance *odfv1alpha1.Stor
 }
 
 func (r *StorageSystemReconciler) isVendorCsvReady(instance *odfv1alpha1.StorageSystem, logger logr.Logger) error {
+	if err := ScaleDownClientOperator(r.Client, instance.Spec.Kind); err != nil {
+		return err
+	}
 
 	csvNames, err := GetVendorCsvNames(r.Client, instance.Spec.Kind)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


leelavg:
fix for https://bugzilla.redhat.com/show_bug.cgi?id=2311857, I tested wrt expectation and scenarios that I could think of, pls review and surface any left out paths orrefactoring efforts. thanks.

iamniting (changes requested):
Hi Leela, Earlier we decided that we will scale it down via the CSV. What is a need of doing it via the deployment now. With this fix we will be scaling down from both the places. IMO we should scale down there itself instead of having a new code.

Here is a PR https://github.com/red-hat-storage/odf-operator/pull/450/files

leelavg:
> What is a need of doing it via the deployment now.
- I explained the reasoning of why deployment is being brought down in code comment and why only setting single CSV replicas to 0 isn't working in PR description. May I know what more info is required?

iamniting:
We had a discussion over a call and we decided to close the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2311857#c7

PR #479, opened 2024-09-19
update ibm odf operator to 1.6.0

iamniting:


Nikhil-Ladha (approved):
/lgtm

hack/make-project-vars.mk
```diff
@@ -5,7 +5,7 @@ ENVTEST_ASSETS_DIR := $(PROJECT_DIR)/testbin
 GOBIN ?= $(BIN_DIR)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-GOPROXY ?= https://proxy.golang.org/
+GOPROXY ?= direct,https://proxy.golang.org
```
malayparida2000, line 8: When we use the direct proxy what exactly it does? Does it slow down the go mod tidy process?

iamniting: yes it does slow down a bit and directly fetches the APIS from the src instead of the goproxy server.

PR #498, opened 2024-10-11
Create a new version of odf-operator

iamniting:


config/manager/manager.yaml
```diff
@@ -0,0 +1,96 @@
...
+  namespace: system
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: odf-operator
+    app.kubernetes.io/managed-by: kustomize
+spec:
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+      app.kubernetes.io/name: odf-operator-webhook
```
leelavg, line 23: why should the operator have webhook as the label value? the app name is `odf-operator` isn't it? iirc pls make sure you aren't editing the selector here if we want the running odf-op be replaces with this as the selector labels are immutable.

cmd/main.go
```diff
@@ -0,0 +1,165 @@
...
+		setupLog.Info("disabling http/2")
+		c.NextProtos = []string{"http/1.1"}
+	}
+
+	tlsOpts := []func(*tls.Config){}
+	if !enableHTTP2 {
+		tlsOpts = append(tlsOpts, disableHTTP2)
+	}
+
+	webhookServer := webhook.NewServer(webhook.Options{
```
leelavg, line 99: what is the port that you are using for webhook here? better to surface (or take from) that in main as well for visibility.

internal/controller/webhook.go
```diff
@@ -0,0 +1,148 @@
...
+				},
+				Operations: []admrv1.OperationType{admrv1.Create},
+			},
+		},
+		SideEffects:             ptr.To(admrv1.SideEffectClassNone),
+		TimeoutSeconds:          ptr.To(int32(30)),
+		AdmissionReviewVersions: []string{"v1"},
+		// fail the validation if webhook can't be reached
+		FailurePolicy: ptr.To(admrv1.Fail),
+	}
```
leelavg, line 62: this whole part is basically a template which doesn't need to be instantiated everytime in a function and so pls move this to a templates package and mutate a deepcopy whenever required in createorupdate method.

internal/controller/webhook.go
```diff
@@ -0,0 +1,148 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	odfwebhook "github.com/red-hat-storage/odf-operator/internal/webhook"
+)
+
+const (
+	WebhookServiceName = "odf-operator-webhook-service"
+	WebhookName        = "csv.odf.openshift.io"
```
leelavg, line 32: are these constants required outside of this package?

internal/controller/webhook.go
```diff
@@ -0,0 +1,148 @@
...
+					Protocol:   corev1.ProtocolTCP,
+					TargetPort: intstr.FromInt32(9443),
+				},
+			},
+			Selector: map[string]string{
+				"app.kubernetes.io/name": "odf-operator-webhook",
+			},
+			Type: corev1.ServiceTypeClusterIP,
+		},
+	}
```
leelavg, line 129: this whole part is basically a template which doesn't need to be instantiated everytime in a function and so pls move this to a templates package and mutate a deepcopy whenever required in createorupdate method.

internal/webhook/csv.go
```diff
@@ -0,0 +1,110 @@
...
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	WebhookPath = "/mutate-operators-coreos-com-v1alpha1-clusterserviceversion"
```
leelavg, line 21: this doesn't need to be GVK identifiable by any means, not user facing, longer names are prone to typos and short name would suffice, how about '/mutate-csv' or can expand csv.

internal/webhook/csv.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+var (
+	ManagedPkgNames = []string{
```
leelavg, line 25: is this required outside of the package?

internal/webhook/csv.go
```diff
@@ -0,0 +1,110 @@
...
+		"mcg-operator",
+		"noobaa-operator",
+		"odf-csi-addons-operator",
+		"csi-addons",
+		"cephcsi-operator",
+		"odf-prometheus-operator",
+		"recipe",
+
+		// extended csvs
+		"ibm-storage-odf-operator",
```
leelavg, line 39: why do we need it here? you can scale down any csv which isn't odf-operator (can be queried from OLM set env easily) in the namespace since we are managing the ns 

the edge case is only operators installed in all namespaces and maybe we need to look at how to filter those out.

(or)

we might want to get this info during build time about all the csvs that we going to manage, wdyt?

internal/webhook/csv.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=create;update;patch
+
+func (r *ClusterServiceVersionDefaulter) Handle(ctx context.Context, req admission.Request) admission.Response {
+	r.context = ctx
+	r.logger = log.FromContext(r.context)
+	r.logger.Info("Request received for ClusterServiceVersion review")
+
+	instance := &opv1a1.ClusterServiceVersion{}
```
leelavg, line 58: pls use the shortname, instance doesn't say anything. same in remaining places.
```suggestion
	csv := &opv1a1.ClusterServiceVersion{}
```

internal/controller/storagesystem_controller.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if err := r.reconcileWebhookService(); err != nil {
+		r.logger.Error(err, "unable to reconcile webhook service")
+		return ctrl.Result{}, err
+	}
+
+	if err := r.reconcileSubscriptionValidatingWebhook(); err != nil {
+		r.logger.Error(err, "unable to register subscription validating webhook")
+		return ctrl.Result{}, err
+	}
```
leelavg, line 60: may I know what else be part of another PR? to name a few as below seems to be related to this PR only
1. watches on webhook and service is missing
2. addition and removal of finalizer on storagesystem is missing
3. removal of webhook and service during uninstallation is missing

internal/controller/storagesystem_controller.go
```diff
@@ -30,27 +31,35 @@ import (
...
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the StorageSystem object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.3/pkg/reconcile
 func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
iamniting: We need to copy-paste the whole reconcile, I will do that as part of the other PR.

internal/controller/storagesystem_controller.go
```diff
@@ -30,27 +31,35 @@ import (
 // StorageSystemReconciler reconciles a StorageSystem object
 type StorageSystemReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme            *runtime.Scheme
+	context           context.Context
```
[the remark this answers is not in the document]
iamniting: I believe it is just a coding style and does not create any impact using either way.

iamniting:
We are making changes to the existing code only

PR #503, opened 2024-11-14
makefile: add target to create a new deps bundle

iamniting:


Makefile
```diff
@@ -170,10 +176,18 @@ undeploy-with-olm: ## Undeploy controller from the K8s cluster
...
+	(git diff --quiet --ignore-matching-lines createdAt bundle/odf-operator && git checkout --quiet bundle/odf-operator) || true
+	(git diff --quiet --ignore-matching-lines createdAt bundle/odf-dependencies && git checkout --quiet bundle/odf-dependencies) || true
 
 .PHONY: bundle
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
+	# Dependencies bundle
+	$(KUSTOMIZE) build config/bundle | $(OPERATOR_SDK) generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) \
+		--output-dir bundle/odf-dependencies --package odf-dependencies
+	$(OPERATOR_SDK) bundle validate bundle/odf-dependencies
+	@mv bundle.Dockerfile bundle.deps.Dockerfile
```
leelavg, line 188: why is this `mv` needed here?

controllers/subscriptions.go
```diff
@@ -111,6 +111,14 @@ func CheckExistingSubscriptions(cli client.Client, desiredSubscription *operator
 	}
 
 	if !subExsist {
+		// Set the catalog source for the odf-dependencies subscription to match that of the odf-operator subscription
+		// This ensures that the odf-dependencies subscription uses the same catalog source across all environments,
+		// including offline and test environments where the catalog name may vary.
```
leelavg, line 116: if this is always the case can we remove loading these specific vars from configmap or env? it's better not to have the data in two places.
(or)
let's take the configmap clubbed with odf-op as source of truth and use that as that'll decouple the hardcoding?

hack/install-odf.sh
```diff
@@ -20,6 +20,15 @@ fi
 
 "$OPERATOR_SDK" run bundle "$BUNDLE_IMG" --timeout=10m --security-context-config restricted -n "$INSTALL_NAMESPACE" --index-image "$CATALOG_DEPS_IMG"
 
+# Check for the presence of the CSVs in the cluster for up to 5 minutes,
+# Since 'oc wait' exits immediately if the resource is not found.
```
leelavg, line 24: probably because of non zero exit code from `oc wait` did you try with `--ignore-node-found`? if we go by dependency chain why not wait for the deps csv status to be successful? sleep is almost always troublesome.

leelavg (commented):
will you be removing the clubbing of other operators info in another PR?

Makefile
iamniting: mv is needed to have 2 different docker files for different bundles.

controllers/subscriptions.go
iamniting: I want to have a place where data is present. In 4.19 we will have something concrete and we can rewrite our logic based on that.

hack/install-odf.sh
iamniting: I will try this though i could not found anything helpful in the `oc wait help`

iamniting:
> will you be removing the clubbing of other operators info in another PR?

Can you pls describe this?

hack/install-odf.sh
iamniting:
```
┌─[nigoyalॐ fedora ~/code/go/src/github.com/red-hat-storage/odf-operator  bundle]
└──╼$ oc wait --timeout=10s --for condition=Available deployment catalog-operatora
Error from server (NotFound): deployments.apps "catalog-operatora" not found

┌─[nigoyalॐ fedora ~/code/go/src/github.com/red-hat-storage/odf-operator  bundle]
└──╼$ oc wait --timeout=10s --for condition=Available deployment catalog-operatora --ignore-not-found
error: unknown flag: --ignore-not-found
See 'oc wait --help' for usage.
```

I tried this, This flag is not available in wait.

Even if we check the deps csv status we have to check for its presence before checking the status. For checking its presence we have to wait and sleep.

Makefile
leelavg: still don't get it, bundle.Dockerfile contains odf-op content and moving it to bundle.deps.Dockerfile doesn't seem correct, both these files are part of this PR?

controllers/subscriptions.go
leelavg: > I want to have a place where data is present.
- then you already have that from configmap isn't it?

> In 4.19 we will have something concrete and we can rewrite our logic based on that.
- sry, my comment isn't related to logic but of data duplication

hack/install-odf.sh
leelavg: ack, resolve.

leelavg:
> > will you be removing the clubbing of other operators info in another PR?
> 
> Can you pls describe this?
- after this PR we'll only be interested in dependent operator channels and CSV isn't it and asked if other operator info, ie, name, namespace, catalog etc will still be required?

iamniting:
> > > will you be removing the clubbing of other operators info in another PR?
> > 
> > 
> > Can you pls describe this?
> 
> * after this PR we'll only be interested in dependent operator channels and CSV isn't it and asked if other operator info, ie, name, namespace, catalog etc will still be required?

Yes it will be required, We will still have to change the channel of the other operators.

Makefile
iamniting: bundle.Dockerfile is created via operator-sdk while running `generate bundle` and we are generating 2 bundles. First one is for deps and second one is for the odf-operator. I am doing a mv on the first bundle.Dockerfile. 

PS: operator-sdk always create the file named bundle.Dockerfile.

controllers/subscriptions.go
iamniting: We had a discussion over a call.

iamniting:
This PR was never LGTM. GitHub shows it as merged because #499 (which was merged) included all its commits with identical hashes.

PR #506, opened 2024-11-15
check CSV phase to be succeeded for container readiness 

leelavg:
odf-operator creates subscription for odf-dependencies and CRDs will become available once respective CSVs are present in the cluster. We check the phase of respective CSVs which are required for odf-op before becoming ready.

Depends on #503

### DIFF
--- a/main.go
+++ b/main.go
@@ -171,7 +171,24 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+
+	// The console plugin depends on CRDs provided by these CSVs.
+	// Checking only the odf-dependencies CSV is insufficient,
+	// because, based on observations, the presence of all required CRDs
+	// is not guaranteed even when the odf-dependencies CSV is in a succeeded state.
+	csvsToBeSuceeded := []string{
+		controllers.OcsSubscriptionStartingCSV,
+		controllers.RookSubscriptionStartingCSV,
+		controllers.NoobaaSubscriptionStartingCSV,
+	}
+	if err := mgr.AddReadyzCheck(
+		"readyz",
+		util.CheckCSVPhase(
+			mgr.GetClient(),
+			operatorNamespace,
+			csvsToBeSuceeded...,
+		),
+	); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -171,7 +171,24 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+
+	// console plugin is dependent on CRDs that are part of these CSVs and we
+	// can't check for only odf-dependencies CSV as OLM doesn't guarantee
+	// (based on observations) existence of all CRDs brought by OLM dependency
+	// mechanism.
```
iamniting, line 178:
``` suggestion
// The console plugin depends on CRDs provided by these CSVs. 
// Checking only the odf-dependencies CSV is insufficient, 
// because, based on observations, the presence of all required CRDs 
// is not guaranteed even when the odf-dependencies CSV is in a succeeded state.
```

pkg/util/readiness.go
```diff
@@ -0,0 +1,56 @@
...
+				if csv.Status.Phase != opv1a1.CSVPhaseSucceeded {
+					return fmt.Errorf("CSV %s is not in Succeeded phase", csv.Name)
+				} else if csv.Status.Phase == opv1a1.CSVPhaseSucceeded {
+					delete(csvMap, csv.Name)
+				}
+			}
+		}
+		if len(csvMap) != 0 {
+			for csvName := range csvMap {
+				return fmt.Errorf("CSV %s is not found", csvName)
```
iamniting, line 51: It will just return only first one. The loop will only run once.

iamniting (changes requested):
Can you pls correct the description of commit, It is still referring to the old logic.

pkg/util/readiness.go
leelavg: that'll still be fine, isn't it? we just aren't ready.

iamniting:
I am adding the screenshot of the observation here for future reference that just checking the odf-dependencies csv is not enough. 

Screenshot from 2024-11-15 16-29-40

pkg/util/readiness.go
iamniting: What i mean to say here is we can just return the first item if the list is not empty, we don't need a loop.

iamniting:
These commits are merged as part of #499. I am closing this.

PR #511, opened 2024-11-21
controllers: Ensure console plugin creation after CSV verification

iamniting:
Delay the creation of the console plugin until after verifying that the required CSVs are present. This ensures the necessary CRDs are present in the cluster, preventing UI errors like "Model does not exist."

### DIFF
--- a/controllers/clusterversion_controller.go
+++ b/controllers/clusterversion_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -37,8 +38,9 @@ import (
 // ClusterVersionReconciler reconciles a ClusterVersion object
 type ClusterVersionReconciler struct {
 	client.Client
-	Scheme      *runtime.Scheme
-	ConsolePort int
+	Scheme            *runtime.Scheme
+	ConsolePort       int
+	OperatorNamespace string
 }
 
 //+kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch;create;update;patch;delete
@@ -58,7 +60,28 @@ func (r *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if err := r.Client.Get(context.TODO(), req.NamespacedName, &instance); err != nil {
 		return ctrl.Result{}, err
 	}
-	if err := r.ensureConsolePlugin(instance.Status.Desired.Version); err != nil {
+
+	if err := r.ensureOdfConsoleConfigMapAndService(); err != nil {
+		logger.Error(err, "Could not ensure configmap and service for odf-console deployment")
+		return ctrl.Result{}, err
+	}
+
+	csvList, err := util.GetNamespaceCSVs(ctx, r.Client, r.OperatorNamespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Check if it is upgrade from 4.17 to 4.18
+	// The new CSVs won't exists while upgrading
+	// They will exists only after new operator has created a new subscription
+	if !util.AreMultipleOdfOperatorCsvsPresent(csvList) {
+		if err := util.ValidateCSVsPresent(csvList, EssentialCSVs...); err != nil {
+			logger.Error(err, "Could not ensure CSVs presence")
+			return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 2}, nil
+		}
+	}
+
+	if err := r.ensureConsolePluginAndCLIDownload(instance.Status.Desired.Version); err != nil {
 		logger.Error(err, "Could not ensure compatibility for ODF consolePlugin")
 		return ctrl.Result{}, err
 	}
@@ -69,12 +92,7 @@ func (r *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 // SetupWithManager sets up the controller with the Manager.
 func (r *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	err := mgr.Add(manager.RunnableFunc(func(context.Context) error {
-		clusterVersion, err := util.DetermineOpenShiftVersion(r.Client)
-		if err != nil {
-			return err
-		}
-
-		return r.ensureConsolePlugin(clusterVersion)
+		return r.ensureOdfConsoleConfigMapAndService()
 	}))
 	if err != nil {
 		return err
@@ -85,14 +103,9 @@ func (r *ClusterVersionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func (r *ClusterVersionReconciler) ensureConsolePlugin(clusterVersion string) error {
+func (r *ClusterVersionReconciler) ensureOdfConsoleConfigMapAndService() error {
 	logger := log.FromContext(context.TODO())
-	// The base path to where the request are sent
-	basePath := console.GetBasePath(clusterVersion)
 	nginxConf := console.NginxConf
-
-	// Customer portal link (CLI Tool download)
-	portalLink := console.CUSTOMER_PORTAL_LINK
 
 	// Get ODF console Deployment
 	odfConsoleDeployment := console.GetDeployment(OperatorNamespace)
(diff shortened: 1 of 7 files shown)


controllers/clusterversion_controller.go
```diff
@@ -68,18 +75,6 @@ func (r *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		if err != nil {
-			return err
-		}
-
-		return r.ensureConsolePlugin(clusterVersion)
-	}))
-	if err != nil {
-		return err
-	}
-
```
SanjalKatiyar, line 82: @bipuladh u remember why this custom runnable was added ?? Removing it seems reasonable, but double checking.

pkg/util/readiness.go
leelavg, line 1: pls recheck if exporting all methods is not necessary.

pkg/util/readiness.go
```diff
@@ -45,24 +43,71 @@ func CheckCSVPhase(c client.Client, namespace string, csvNames ...string) health
...
+	return nil
+}
+
+func GetCSVByName(csvList *opv1a1.ClusterServiceVersionList, name string) (*opv1a1.ClusterServiceVersion, bool) {
+
+	for _, csv := range csvList.Items {
+		if csv.Name == name {
+			return &csv, true
+		}
+	}
```
leelavg, line 102: pls query by index, here there is a creation of new var and it creates many not required values escaping the function scope.

controllers/clusterversion_controller.go
```diff
@@ -58,7 +60,18 @@ func (r *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		return ctrl.Result{}, err
 	}
-	if err := r.ensureConsolePlugin(instance.Status.Desired.Version); err != nil {
+
+	if err := r.ensureOdfConsoleConfigMapAndService(); err != nil {
+		logger.Error(err, "Could not ensure configmap and service for odf-console deployment")
+		return ctrl.Result{}, err
+	}
+
+	if err := util.EnsureCSVsPresent(ctx, r.Client, r.OperatorNamespace, EssentialCSVs...); err != nil {
```
leelavg, line 69: if this is all we want, why to continue to use readiness probe involving a hack?

controllers/clusterversion_controller.go
```diff
@@ -58,7 +60,18 @@ func (r *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		logger.Error(err, "Could not ensure CSVs in Succeeded state")
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 2}, nil
```
leelavg, line 71: 2 sec is demanding, the observed time was 60s in total and 10s interval will be relaxing. (or) remove the time and add a csv predicate creation event w/ essential CSVs.

controllers/subscription_controller.go
```diff
@@ -290,6 +291,22 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		odfDepsSub := GetStorageClusterSubscriptions()[0]
+
+		if odfDepsSub.Spec.Package != OdfDepsSubscriptionPackage {
+			panic("odfDepsSub variable is expected to contain the 'odf-dependencies' subscription. " +
+				"Ensure the 'odf-dependencies' subscription indexed at 0.")
+		}
+
+		return EnsureDesiredSubscription(r.Client, odfDepsSub)
+
```
leelavg, line 304: pls remove extra lines, all actions seems to be a single unit.

controllers/subscription_controller.go
```diff
@@ -290,6 +291,22 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
```
leelavg, line 294: may I know what's the expectation w/ runnablefunc?

controllers/subscription_controller.go
```diff
@@ -290,6 +291,22 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+		odfDepsSub := GetStorageClusterSubscriptions()[0]
+
+		if odfDepsSub.Spec.Package != OdfDepsSubscriptionPackage {
+			panic("odfDepsSub variable is expected to contain the 'odf-dependencies' subscription. " +
+				"Ensure the 'odf-dependencies' subscription indexed at 0.")
```
leelavg, line 300: who needs to ensure this? if this is static at build time why not move this to main? may I know what's the requirement? why can't we make an arrangement for caller to explicitly get deps subscription and then the remaining?

controllers/clusterversion_controller.go
iamniting: We need both of these. The readiness probe will stop the customer from creating the storage cluster. Creating the console plugin later will show the popup once everything is ready.

controllers/clusterversion_controller.go
iamniting: I don't want to add long intervals, Adding a longer interval can delay the creation of the web console, which means a delay in the installation. Also, this will happen only when there is an installation or upgrade. It reminds me to handle the upgrade case.

controllers/subscription_controller.go
iamniting: We developer needs to ensure this while coding. I added a panic so that we can catch this while testing. Maybe I can create a unit test to ensure this.

controllers/clusterversion_controller.go
leelavg: > I don't want to add long intervals,
- not sure for what qualifies as longer or shorter interval in a reconciliation and so dropping my comment.

controllers/clusterversion_controller.go
iamniting: It will stop on the UI.

pkg/util/readiness.go
iamniting: Done

pkg/util/readiness.go
iamniting: Done

controllers/subscription_controller.go

controllers/subscription_controller.go

leelavg (approved):
/hold
Unhold if no one else is going to review. Thanks.

bipuladh (approved):
Looks good to me.

controllers/clusterversion_controller.go
```diff
@@ -58,7 +60,28 @@ func (r *ClusterVersionReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	csvList, err := util.GetNamespaceCSVs(ctx, r.Client, r.OperatorNamespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Check if it is upgrade from 4.17 to 4.18
+	// The new CSVs won't exists while upgrading
+	// They will exists only after new operator has created a new subscription
+	if !util.AreMultipleOdfOperatorCsvsPresent(csvList) {
```
SanjalKatiyar, line 77: just a general comment/doubt (no need to change the current implementation):

this controller will only run when either OCP is upgraded (i.e., For(&configv1.ClusterVersion{})) or when controller will restart after ODF is upgraded (in which case new CSVs will be existing soon)... we don't really need `!util.AreMultipleOdfOperatorCsvsPresent(csvList)` check, right ?? or am I missing any case ??

iamniting: This check is to make sure that we don't block the reconcile in case of upgrades. For example assume customers has messed up the catalog for some reason in that case only odf will be upgraded and other operators wont be upgraded. So we should not block the console updation in that case right?

SanjalKatiyar: > So we should not block the console updation in that case right?

make sense.

> For example assume customers has messed up the catalog for some reason in that case only odf will be upgraded and other operators wont be upgraded.

If ODF is upgraded and others won't, then we are still blocked right ?? because `!util.AreMultipleOdfOperatorCsvsPresent(csvList)` will be `true` in that case.

iamniting: Yes, that's right, It will skip if the odf-operator upgrade is in progress.

SanjalKatiyar: actually that was my initial doubt, why would this code be executed when ODF upgrade is in progress: https://github.com/red-hat-storage/odf-operator/pull/511#discussion_r1875486724

iamniting: Once the manager is started, The controllers will also get triggered. 

I have added this check to keep the behavior as before and that is once the odf-operator is upgraded, I also upgrade the console plugin.

iamniting:
This PR is not required as the console will be adding the CRDS check and loading screen.

PR #516, opened 2025-01-15
Create a CSV webhook

iamniting:

### DIFF
--- a/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
@@ -161,6 +161,16 @@ spec:
         - apiGroups:
           - admissionregistration.k8s.io
           resources:
+          - mutatingwebhookconfigurations
+          verbs:
+          - create
+          - get
+          - list
+          - update
+          - watch
+        - apiGroups:
+          - admissionregistration.k8s.io
+          resources:
           - validatingwebhookconfigurations
           verbs:
           - delete
@@ -421,6 +431,10 @@ spec:
                   initialDelaySeconds: 15
                   periodSeconds: 20
                 name: manager
+                ports:
+                - containerPort: 9443
+                  name: webhook-server
+                  protocol: TCP
                 readinessProbe:
                   httpGet:
                     path: /readyz
@@ -441,6 +455,10 @@ spec:
                     drop:
                     - ALL
                   readOnlyRootFilesystem: true
+                volumeMounts:
+                - mountPath: /tmp/k8s-webhook-server/serving-certs
+                  name: cert
+                  readOnly: true
               securityContext:
                 runAsNonRoot: true
               serviceAccountName: odf-operator-controller-manager
@@ -450,6 +468,11 @@ spec:
                 key: node.ocs.openshift.io/storage
                 operator: Equal
                 value: "true"
+              volumes:
+              - name: cert
+                secret:
+                  defaultMode: 420
+                  secretName: odf-operator-webhook-server-cert
       - label:
           app: odf-console
         name: odf-console
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -25,6 +25,7 @@ namePrefix: odf-operator-
 # endpoint w/o any authn/z, please comment the following line.
 patchesStrategicMerge:
 - manager_auth_proxy_patch.yaml
+- manager_webhook_patch.yaml
 
 # Mount the controller config file for loading manager configurations
 # through a ComponentConfig type
--- a/config/manager/webhook_service.yaml
+++ b/config/manager/webhook_service.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: odf-operator-webhook-server-cert
+  name: webhook-server-service
+  namespace: system
+spec:
+  ports:
+  - name: https
+    port: 443
+    protocol: TCP
+    targetPort: 9443
+  selector:
+    app.kubernetes.io/name: odf-operator
--- a/controllers/storagesystem_controller.go
+++ b/controllers/storagesystem_controller.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"go.uber.org/multierr"
+	admrv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -86,7 +87,7 @@ func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	metrics.ReportODFSystemMapMetrics(instance.Name, instance.Spec.Name, instance.Spec.Namespace, string(instance.Spec.Kind))
 
 	// Reconcile changes
-	result, reconcileError := r.reconcile(instance, logger)
+	result, reconcileError := r.reconcile(ctx, instance, logger)
 
 	// Apply status changes
 	statusError := r.Client.Status().Update(context.TODO(), instance)
@@ -105,7 +106,7 @@ func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 }
 
-func (r *StorageSystemReconciler) reconcile(instance *odfv1alpha1.StorageSystem, logger logr.Logger) (ctrl.Result, error) {
+func (r *StorageSystemReconciler) reconcile(ctx context.Context, instance *odfv1alpha1.StorageSystem, logger logr.Logger) (ctrl.Result, error) {
 
 	var err error
 
@@ -156,6 +157,10 @@ func (r *StorageSystemReconciler) reconcile(instance *odfv1alpha1.StorageSystem,
 		return ctrl.Result{}, err
 	}
 
+	if ok, err := isCsvWebhookPresent(ctx, r.Client, logger); !ok {
+		return ctrl.Result{}, err
+	}
+
 	err = r.ensureSubscriptions(instance, logger)
 	if err != nil {
 		return ctrl.Result{}, err
@@ -197,6 +202,19 @@ func (r *StorageSystemReconciler) validateStorageSystemSpec(instance *odfv1alpha
 	}
 
 	return nil
+}
+
+func isCsvWebhookPresent(ctx context.Context, cli client.Client, logger logr.Logger) (bool, error) {
+
+	whConfig := &admrv1.MutatingWebhookConfiguration{}
+	key := client.ObjectKey{Name: csvWebhook.Name}
+
+	if err := cli.Get(ctx, key, whConfig); err != nil {
+		logger.Error(err, "waiting for csv webhook to be created")
+		return false, err
+	}
+
+	return true, nil
 }
 
 func (r *StorageSystemReconciler) ensureSubscriptions(instance *odfv1alpha1.StorageSystem, logger logr.Logger) error {
--- a/controllers/subscription_controller.go
+++ b/controllers/subscription_controller.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"go.uber.org/multierr"
+	admrv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -47,6 +48,7 @@ type SubscriptionReconciler struct {
 	client.Client
 	Scheme            *runtime.Scheme
 	Recorder          *EventReporter
+	OperatorNamespace string
 	ConditionName     string
 	OperatorCondition conditions.Condition
 }
@@ -57,6 +59,7 @@ type SubscriptionReconciler struct {
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=installplans,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions/finalizers,verbs=update
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=operatorconditions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;create;update
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
@@ -75,13 +78,16 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return ctrl.Result{}, err
 	}
 
-	err = r.setOperatorCondition(logger, req.NamespacedName.Namespace)
+	err = r.setOperatorCondition(logger, r.OperatorNamespace)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	err = r.ensureSubscriptions(logger, req.NamespacedName.Namespace)
-	if err != nil {
+	if err = reconcileCsvWebhook(ctx, r.Client, logger, r.OperatorNamespace); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err = r.ensureSubscriptions(logger, r.OperatorNamespace); err != nil {
 		return ctrl.Result{}, err
 	}
 
@@ -290,12 +296,22 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	enqueueOdfSub := handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
+			req := reconcile.Request{}
+			req.Name = "odf-operator"
+			req.Namespace = r.OperatorNamespace
+			return []reconcile.Request{req}
+		},
+	)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&operatorv1alpha1.Subscription{},
 			builder.WithPredicates(generationChangedPredicate, subscriptionPredicate)).
 		Owns(&operatorv1alpha1.Subscription{},
 			builder.WithPredicates(generationChangedPredicate)).
 		Watches(&operatorv2.OperatorCondition{}, enqueueFromCondition, builder.WithPredicates(conditionPredicate)).
+		Watches(&admrv1.MutatingWebhookConfiguration{}, enqueueOdfSub, builder.WithPredicates(generationChangedPredicate)).
 		Complete(r)
 }
 
--- a/controllers/webhook.go
+++ b/controllers/webhook.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2025 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	admrv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	CsvWebhookPath = "/mutate-operators-coreos-com-v1alpha1-csv"
+)
+
+var (
+	csvWebhookService = corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "odf-operator-webhook-server-service",
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Name:       "https",
+					Port:       443,
+					Protocol:   corev1.ProtocolTCP,
+					TargetPort: intstr.FromInt32(9443),
+				},
+			},
+			Selector: map[string]string{
+				"app.kubernetes.io/name": "odf-operator",
+			},
+			Type: corev1.ServiceTypeClusterIP,
+		},
+	}
+
+	csvWebhook = admrv1.MutatingWebhook{
+		Name: "csv.odf.openshift.io",
+		ClientConfig: admrv1.WebhookClientConfig{
+			Service: &admrv1.ServiceReference{
+				Name: csvWebhookService.Name,
+				Path: ptr.To(CsvWebhookPath),
+				Port: ptr.To(int32(443)),
+			},
+		},
+		Rules: []admrv1.RuleWithOperations{{
+			Rule: admrv1.Rule{
+				APIGroups:   []string{"operators.coreos.com"},
+				APIVersions: []string{"v1alpha1"},
+				Resources:   []string{"clusterserviceversions"},
+				Scope:       ptr.To(admrv1.NamespacedScope),
+			},
+			Operations: []admrv1.OperationType{admrv1.Create},
+		}},
+		SideEffects:             ptr.To(admrv1.SideEffectClassNone),
+		TimeoutSeconds:          ptr.To(int32(30)),
+		AdmissionReviewVersions: []string{"v1"},
+		// fail the admission if webhook can't be reached
+		FailurePolicy: ptr.To(admrv1.Fail),
+	}
+)
+
+func reconcileCsvWebhook(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	if err := reconcileCsvMutatingWebhookService(ctx, cli, logger, operatorNamespace); err != nil {
+		logger.Error(err, "unable to reconcile webhook service")
+		return err
+	}
+
+	if err := reconcileCsvMutatingWebhookConfiguration(ctx, cli, logger, operatorNamespace); err != nil {
+		logger.Error(err, "unable to register csv mutating webhook")
+		return err
+	}
+
+	return nil
+}
+
+func reconcileCsvMutatingWebhookService(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	svc := &corev1.Service{}
+	svc.Name = csvWebhookService.Name
+	svc.Namespace = operatorNamespace
+
+	res, err := controllerutil.CreateOrUpdate(ctx, cli, svc, func() error {
+		if svc.Annotations == nil {
+			svc.Annotations = map[string]string{}
+		}
+		svc.Annotations["service.beta.openshift.io/serving-cert-secret-name"] = "odf-operator-webhook-server-cert"
+		csvWebhookService.Spec.DeepCopyInto(&svc.Spec)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	logger.Info("successfully created or updated webhook service", "operation", res, "name", svc.Name)
+	return nil
+}
+
+func reconcileCsvMutatingWebhookConfiguration(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	whConfig := &admrv1.MutatingWebhookConfiguration{}
+	whConfig.Name = csvWebhook.Name
+
+	res, err := controllerutil.CreateOrUpdate(ctx, cli, whConfig, func() error {
+
+		if whConfig.Annotations == nil {
+			whConfig.Annotations = map[string]string{}
+		}
+		// openshift fills in the ca on finding this annotation
+		whConfig.Annotations["service.beta.openshift.io/inject-cabundle"] = "true"
+
+		var caBundle []byte
+		if len(whConfig.Webhooks) == 0 {
+			whConfig.Webhooks = make([]admrv1.MutatingWebhook, 1)
+		} else {
+			// do not mutate CA bundle that was injected by openshift
+			caBundle = whConfig.Webhooks[0].ClientConfig.CABundle
+		}
+
+		// webhook desired state
+		wh := &whConfig.Webhooks[0]
+		csvWebhook.DeepCopyInto(wh)
+
+		wh.Name = whConfig.Name
+		// only send requests received from own namespace
+		wh.NamespaceSelector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"kubernetes.io/metadata.name": operatorNamespace,
+			},
+		}
+
+		wh.ObjectSelector = &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      "olm.copiedFrom",
+					Operator: metav1.LabelSelectorOpDoesNotExist,
+				},
+				{
+					Key:      fmt.Sprintf(CsvLabelKey, "odf-operator", operatorNamespace),
+					Operator: metav1.LabelSelectorOpDoesNotExist,
+				},
+			},
+		}
+
+		/*// create cel rules
+		var conditions []string
+		for _, pkg := range PkgNames {
+			labelKey := fmt.Sprintf(CsvLabelKey, pkg, operatorNamespace)
+			conditions = append(conditions, fmt.Sprintf(`object.metadata.labels.contains("%s")`, labelKey))
+		}
+
+		wh.MatchConditions = []admrv1.MatchCondition{{
+			Name:       "has_at_least_one_target_label",
+			Expression: strings.Join(conditions, " || "),
+		}}*/
+
+		// preserve the existing (injected) CA bundle if any
+		wh.ClientConfig.CABundle = caBundle
+		// send request to the service running in own namespace
+		wh.ClientConfig.Service.Namespace = operatorNamespace
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	logger.Info("successfully created or updated webhook", "operation", res, "name", whConfig.Name)
+	return nil
+}
--- a/webhook/csv.go
+++ b/webhook/csv.go
@@ -1,0 +1,147 @@
+/*
+Copyright 2025 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package webhook
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/go-logr/logr"
+	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/red-hat-storage/odf-operator/controllers"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type ClusterServiceVersionDeploymentScaler struct {
+	client.Client
+
+	Decoder           admission.Decoder
+	OperatorNamespace string
+}
+
+// +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;patch
+
+func (r *ClusterServiceVersionDeploymentScaler) Handle(ctx context.Context, req admission.Request) admission.Response {
+
+	logger := log.FromContext(ctx)
+	logger.Info("request received for csv review")
+
+	csv := &opv1a1.ClusterServiceVersion{}
+	if err := r.Decoder.Decode(req, csv); err != nil {
+		logger.Error(err, "failed decoding admission review as csv")
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed decoding admission review as csv: %v", err))
+	}
+
+	if ok := r.isCsvManagedByOdf(csv); !ok {
+		logger.Info("ignoring csv as it is not a csv managed by ODF")
+		return admission.Allowed("csv is not managed by ODF")
+	}
+
+	running, err := r.isPreviousCsvHasRunningDeployments(ctx, logger, csv)
+	if err != nil {
+		logger.Error(err, "failed getting replicas from csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed getting replicas from csv: %v", err))
+	}
+
+	if running {
+		logger.Info("ignoring csv as the previous csv deployments are running")
+		return admission.Allowed("previous csv deployments are running")
+	}
+
+	r.scaleDownCsvDeployments(logger, csv)
+
+	marshaledCsv, err := json.Marshal(csv)
+	if err != nil {
+		logger.Error(err, "failed marshaling csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed marshaling csv: %v", err))
+	}
+
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledCsv)
+}
+
+func (r *ClusterServiceVersionDeploymentScaler) isPreviousCsvHasRunningDeployments(ctx context.Context, logger logr.Logger, csv *opv1a1.ClusterServiceVersion) (bool, error) {
+
+	if csv.Spec.Replaces == "" {
+		logger.Info("csv.Spec.Replaces is not populated")
+		return false, nil
+	}
+
+	prevCsv := &opv1a1.ClusterServiceVersion{}
+	key := client.ObjectKey{Name: csv.Spec.Replaces, Namespace: csv.Namespace}
+
+	if err := r.Client.Get(ctx, key, prevCsv); errors.IsNotFound(err) {
+		// new install where an previous csv does not exists
+		return false, nil
+	} else if err != nil {
+		logger.Error(err, "failed getting previous csv")
+		return false, err
+	}
+
+	deployments := prevCsv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+	for i := range deployments {
+		deployment := &deployments[i]
+		if *deployment.Spec.Replicas > 0 {
+			// upgrade case where an older csv is found with replica 1
+			return true, nil
+		}
+	}
+
+	// upgrade case where an older csv is found with replica 0
+	return false, nil
+}
+
+func (r *ClusterServiceVersionDeploymentScaler) scaleDownCsvDeployments(logger logr.Logger, csv *opv1a1.ClusterServiceVersion) {
+
+	logger.Info("mutating requested csv")
+
+	for i := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec.Replicas = ptr.To(int32(0))
+	}
+}
+
+func (r *ClusterServiceVersionDeploymentScaler) isCsvManagedByOdf(csv *opv1a1.ClusterServiceVersion) bool {
+
+	if csv.Namespace != r.OperatorNamespace {
+		return false
+	}
+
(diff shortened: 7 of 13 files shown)


config/default/kustomization.yaml
```diff
@@ -25,6 +25,7 @@ namePrefix: odf-operator-
 # endpoint w/o any authn/z, please comment the following line.
 patchesStrategicMerge:
 - manager_auth_proxy_patch.yaml
+- manager_webhook_patch.yaml
```
leelavg, line 28: why is this required?

controllers/webhook.go
```diff
@@ -0,0 +1,166 @@
...
+				},
+				Operations: []admrv1.OperationType{admrv1.Create},
+			},
+		},
+		SideEffects:             ptr.To(admrv1.SideEffectClassNone),
+		TimeoutSeconds:          ptr.To(int32(30)),
+		AdmissionReviewVersions: []string{"v1"},
+		// fail the validation if webhook can't be reached
+		FailurePolicy: ptr.To(admrv1.Fail),
+	}
```
leelavg, line 80: this is recreated in every func invocation, why can't move this to a template and deep copy whatever is required?

controllers/webhook.go
```diff
@@ -0,0 +1,166 @@
...
+			Selector: map[string]string{
+				"app.kubernetes.io/name": "odf-operator",
+			},
+			Type: corev1.ServiceTypeClusterIP,
+		},
+	}
+
+	svc := &corev1.Service{
+		ObjectMeta: webhookService.ObjectMeta,
+	}
```
leelavg, line 151: this is recreated in every func invocation, why can't move this to a template and deep copy whatever is required?

config/default/kustomization.yaml
iamniting: added as a variable.

controllers/webhook.go
iamniting: added as a variable.

controllers/webhook.go
iamniting: added as a variable.

leelavg:
does this PR needs a push? where are you registering the webhook?

iamniting:
> does this PR needs a push? where are you registering the webhook?

No it is getting created in `controllers/storagesystem_controller.go` and `controllers/subscription_controller.go`

leelavg:
> > does this PR needs a push? where are you registering the webhook?
> 
> No it is getting created in `controllers/storagesystem_controller.go` and `controllers/subscription_controller.go`
- trying one more time, you have a `type` created in `webhook/csv.go` which has `SetupWebhookWithManager` method defined and I'm unable to find where this method was called.

iamniting:
> > > does this PR needs a push? where are you registering the webhook?
> > 
> > 
> > No it is getting created in `controllers/storagesystem_controller.go` and `controllers/subscription_controller.go`
> 
> * trying one more time, you have a `type` created in `webhook/csv.go` which has `SetupWebhookWithManager` method defined and I'm unable to find where this method was called.

I fixed it, It was missed, It is called in the main.

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	marshaledInstance, err := json.Marshal(instance)
+	if err != nil {
+		r.log.Error(err, "failed marshaling csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed marshaling csv: %v", err))
+	}
+
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledInstance)
+}
+
+func (r *ClusterServiceVersionDefaulter) getOldCsvReplicas(instance *opv1a1.ClusterServiceVersion) (int32, error) {
```
leelavg, line 83: rather than `instance`, `csv` is more discoverable as param var.

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (r *ClusterServiceVersionDefaulter) isOdfManagedCsv(instance *opv1a1.ClusterServiceVersion) bool {
+
+	if instance.Namespace != r.OperatorNamespace {
+		return false
+	}
+
+	for _, pkgName := range controllers.PkgNames {
+		if strings.HasPrefix(instance.ObjectMeta.Name, pkgName) {
```
leelavg, line 128: ObjectMeta and TypeMeta is embedded into the struct and could directly use `.Name`

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+			// upgrade case where an older csv is found with replica 1
+			return 1, nil
+		}
+	}
+
+	// upgrade case where an older csv is found with replica 0
+	return 0, nil
+}
+
+func (r *ClusterServiceVersionDefaulter) mutateCsv(instance *opv1a1.ClusterServiceVersion, replicas *int32) {
```
leelavg, line 113: why do we need a func here, this is neither a single unit nor being called multiple times.

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+		r.log.Error(err, "failed marshaling csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed marshaling csv: %v", err))
+	}
+
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledInstance)
+}
+
+func (r *ClusterServiceVersionDefaulter) getOldCsvReplicas(instance *opv1a1.ClusterServiceVersion) (int32, error) {
+
+	if instance.Spec.Replaces == "" {
```
leelavg, line 85: so this is a new operator from this release and should we also check if we have the name/package in our records or else this will always stay scaled down?

bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
```diff
@@ -158,6 +158,16 @@ spec:
...
+        - apiGroups:
+          - admissionregistration.k8s.io
+          resources:
+          - mutatingwebhookconfigurations
+          verbs:
+          - create
+          - get
+          - list
+          - update
+          - watch
```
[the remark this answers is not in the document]
iamniting: Who will delete this? We can not even have an owner ref added to the webhook.

controllers/subscription_controller.go
```diff
@@ -75,28 +81,33 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	}
 
-	err = r.setOperatorCondition(logger, req.NamespacedName.Namespace)
+	err = r.setOperatorCondition(logger)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	err = r.ensureSubscriptions(logger, req.NamespacedName.Namespace)
+	err = ensureWebhook(r.ctx, r.Client, logger, r.OperatorNamespace)
```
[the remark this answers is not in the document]
iamniting: Subscriptions are handled from these 2 controllers. any of these can be triggered first, and we need to make sure we have the webhook before subscriptions.

controllers/webhook.go
```diff
@@ -0,0 +1,191 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	WebhookServiceName = "odf-operator-webhook-service"
+	WebhookName        = "csv.odf.openshift.io"
+	WebhookPath        = "/mutate-operators-coreos-com-v1alpha1-csv"
```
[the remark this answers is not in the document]
iamniting: only webhook path is used, I wil unexport others.

webhook/csv.go
iamniting: i would prefer instance, because that is being used and created by scaffolding mostly at controller layers.

webhook/csv.go
iamniting: If it is a new package, we should scale down by default, We should scale it up by the controller if needed.

bundle/odf-operator/manifests/odf-operator-webhook-service_v1_service.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: webhook-server-cert
```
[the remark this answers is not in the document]
iamniting: renamed it to `odf-operator-webhook-server-cert`

controllers/storagesystem_controller.go
```diff
@@ -261,12 +274,38 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
+	enqueueStorageSystemRequest := handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
+			logger := log.FromContext(ctx)
+			ssList := &odfv1alpha1.StorageSystemList{}
+			err := r.Client.List(ctx, ssList, &client.ListOptions{Namespace: r.OperatorNamespace})
```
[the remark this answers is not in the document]
iamniting: Done

controllers/webhook.go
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
iamniting: Done

controllers/webhook.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+var (
```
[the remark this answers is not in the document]
iamniting: What will be the benefits of that?

controllers/subscription_controller.go
```diff
@@ -290,12 +301,40 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				return []reconcile.Request{}
+			}
+			return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: sub.Name, Namespace: sub.Namespace}}}
+		},
+	)
+
+	err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+
+		logger := log.FromContext(ctx)
+		err := ensureWebhook(ctx, r.Client, logger, r.OperatorNamespace)
```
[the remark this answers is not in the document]
iamniting: It should not cause any issues i believe, This was created so that even if we hit other issues, we still create the webhook and odf-deps subs.

bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
leelavg: can add an gh issue tracker and close this comment.

controllers/operatorscaler_controller.go
```diff
@@ -93,6 +93,13 @@ var (
 		},
 	}
 
+	PkgNames = func() (pkgNames []string) {
```
leelavg, line 96: wasn't this part of merged PR 🤔?

controllers/webhook.go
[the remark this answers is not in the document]
leelavg: to me it says what you are reconciling, in this file I can see the name of the resource is being taken from this object which is a copy of another var.

atleast for the selector you are spreading the info into two places, I read template for adding/extending new info rather than overwriting or spreading content into two places.

nevertheless I don't have a reason to put into words asking for a mandatory change.

webhook/csv.go
leelavg: sry, I don't see such a recommendation/scaffolding looking at https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata (maybe I'm missing legit sources? open to be corrected) setting that aside, if the param is passed outside of reconciler to a function what would `instance` mean in that function context?

webhook/csv.go
leelavg: _what and where changes are made to the csv._ .... if multiple mutations are being done or being called from multiple places which isn't the case here and inlining makes more sense to me.

leelavg (commented):
can resolve the comments which I already ack'd (~👍🏼), thanks.

bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
iamniting: Created it #556

webhook/csv.go
iamniting: Currently, we have only one change. Although it is not being called from multiple places, it's better to use a function to maintain consistency in the webhook workflow. Also we never know if this can grow in future.

webhook/csv.go
iamniting: changed it to `csv`

controllers/webhook.go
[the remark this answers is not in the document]
iamniting: I have removed the namespace selector from the variable.

controllers/subscription_controller.go
```diff
@@ -290,12 +301,40 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		if err != nil {
+			return err
+		}
+
+		odfDepsSub := GetStorageClusterSubscriptions()[0]
```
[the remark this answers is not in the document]
iamniting: I think this can be resolved, as this code is removed.

controllers/subscription_controller.go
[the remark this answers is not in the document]
iamniting: This is removed. Can be resolved.

controllers/storagesystem_controller.go
```diff
@@ -47,9 +51,11 @@ const (
...
 type StorageSystemReconciler struct {
 	client.Client
-	Log      logr.Logger
-	Scheme   *runtime.Scheme
-	Recorder *EventReporter
+	Log               logr.Logger
+	Scheme            *runtime.Scheme
+	Recorder          *EventReporter
+	OperatorNamespace string
+	ctx               context.Context
```
nb-ohad, line 58: `ctx` is a pair request variable and should not be stored on the reconciler object which is shared between event. 

Bugs will only be visible when we will set up the controller concurrency to more then 1, but even if we never will, this is still a bug. As a matter of fact the person which will raise the concurrency will not even know that they are introducing a bug, as reproduction of the but will be depended on exact scheduling conditions. 

For data the is request scoped, please pass it through the function invocation as parameters (as separate values or as a compound struct), or alternatively follow the approach taken here: https://github.com/ceph/ceph-csi-operator/blob/main/internal/controller/driver_controller.go#L87-L101

controllers/storagesystem_controller.go
```diff
@@ -64,11 +70,14 @@ type StorageSystemReconciler struct {
...
 //+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch;create;update
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=delete
+//+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;create;update
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
 func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := r.Log.WithValues("instance", req.NamespacedName)
+
+	r.ctx = ctx
```
nb-ohad, line 79: Please take a look at the last comment
```suggestion
	r.ctx = ctx
```

controllers/storagesystem_controller.go
```diff
@@ -156,6 +165,11 @@ func (r *StorageSystemReconciler) reconcile(instance *odfv1alpha1.StorageSystem,
 		return ctrl.Result{}, err
 	}
 
+	err = ensureWebhook(r.ctx, r.Client, logger, r.OperatorNamespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
nb-ohad, line 172: We should not be ensuring the webhook from this controller. You can validate that webhook is installed but only a single controller should be responsible to set up and reconcile the webhook.

controllers/storagesystem_controller.go
```diff
@@ -261,12 +275,38 @@ func (r *StorageSystemReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+					NamespacedName: types.NamespacedName{
+						Namespace: ss.Namespace,
+						Name:      ss.Name,
+					},
+				})
+			}
+
+			return request
+		},
+	)
```
nb-ohad, line 301: Why are we adding code if we are getting rid of storage system?

controllers/subscription_controller.go
```diff
@@ -47,8 +48,10 @@ type SubscriptionReconciler struct {
 	client.Client
 	Scheme            *runtime.Scheme
 	Recorder          *EventReporter
+	OperatorNamespace string
 	ConditionName     string
 	OperatorCondition conditions.Condition
+	ctx               context.Context
```
nb-ohad, line 54: See previous comment of why this is a bad practice and should be avoided

controllers/subscription_controller.go
```diff
@@ -57,11 +60,13 @@ type SubscriptionReconciler struct {
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=installplans,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions/finalizers,verbs=update
 //+kubebuilder:rbac:groups=operators.coreos.com,resources=operatorconditions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;create;update
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.3/pkg/reconcile
 func (r *SubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 
+	r.ctx = ctx
```
nb-ohad, line 69:
```suggestion
```

controllers/subscription_controller.go
```diff
@@ -290,12 +300,25 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	enqueueOdfSub := handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
+			logger := log.FromContext(ctx)
+			sub, err := GetOdfSubscription(r.Client)
```
nb-ohad, line 306: We should not be running client command inside an Enqueue or request or a predicate. 
The only 2 ways we are allowed to invoke is: 
- Cast the obj to the event's object then use information on it to get to the desired CR (maybe using an ownerref, or an annotation)
- Locate the object before setting defining setting the enqueue command and use that name as a constant within the enqueue function

If we need to load the name of a constant CR we should do that before setting the E

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		SideEffects:             ptr.To(admrv1.SideEffectClassNone),
+		TimeoutSeconds:          ptr.To(int32(30)),
+		AdmissionReviewVersions: []string{"v1"},
+		// fail the admission if webhook can't be reached
+		FailurePolicy: ptr.To(admrv1.Fail),
+	}
+)
+
+func ensureWebhook(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
```
nb-ohad, line 93: Please use the proper terminology here: `reconcile` not `ensure`. And be more specific with the name:

```suggestion
func reconicCsvWebhook(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	webhookServiceName = "odf-operator-webhook-service"
```
nb-ohad, line 33: We might want to have different webhooks in the future, so let's be more specific with our naming
```suggestion
	csvWebhookServiceName = "odf-operator-csv-webhook-service"
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if err := reconcileCsvMutatingWebhookService(ctx, cli, logger, operatorNamespace); err != nil {
+		logger.Error(err, "unable to reconcile webhook service")
+		return err
+	}
+
+	if err := reconcileCsvMutatingWebhookConfiguration(ctx, cli, logger, operatorNamespace); err != nil {
+		logger.Error(err, "unable to register csv mutating webhook")
+		return err
+	}
```
nb-ohad, line 103: Why aren't these defined and called from within the file of the controller that sets up/reconcile the webhook?
Why do we need a separate file for this? a separate function?

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func reconcileCsvMutatingWebhookService(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	svc := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      webhookService.ObjectMeta.Name,
+			Namespace: operatorNamespace,
+		},
+	}
```
nb-ohad, line 115: Just a suggestion to help with readability:
```suggestion
	svc := &corev1.Service{}
	svc.Name =  webhookService.ObjectMeta.Name
	svc.Namespace = operatorNamespace
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return nil
+}
+
+func reconcileCsvMutatingWebhookService(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	svc := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      webhookService.ObjectMeta.Name,
```
nb-ohad, line 112: You have a constant, use it:
```suggestion
			Name:      webhookServiceName,
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+					Protocol:   corev1.ProtocolTCP,
+					TargetPort: intstr.FromInt32(9443),
+				},
+			},
+			Selector: map[string]string{
+				"app.kubernetes.io/name": "odf-operator",
+			},
+			Type: corev1.ServiceTypeClusterIP,
+		},
+	}
```
nb-ohad, line 57: You are only interested in the spec, there is no point in having a complete object here
```suggestion
	webhookService = corev1.ServiceSpec{
		Ports: []corev1.ServicePort{
			{
				Name:       "https",
				Port:       443,
				Protocol:   corev1.ProtocolTCP,
				TargetPort: intstr.FromInt32(9443),
			},
		},
		Selector: map[string]string{
			"app.kubernetes.io/name": "odf-operator",
		},
		Type: corev1.ServiceTypeClusterIP,
	}
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+
+	logger.Info("successfully created or updated webhook service", "operation", res, "name", svc.Name)
+	return nil
+}
+
+func reconcileCsvMutatingWebhookConfiguration(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	whConfig := &admrv1.MutatingWebhookConfiguration{}
+	whConfig.Name = csvWebhook.Name
```
nb-ohad, line 135: You have a constant, use it:
```suggestion
	whConfig.Name = webhookName
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+			},
+			Selector: map[string]string{
+				"app.kubernetes.io/name": "odf-operator",
+			},
+			Type: corev1.ServiceTypeClusterIP,
+		},
+	}
+
+	csvWebhook = admrv1.MutatingWebhook{
+		Name: webhookName,
```
nb-ohad, line 60: There is no point in having the name stick to the template. You already have the name defined as a constant
```suggestion
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+	whConfig := &admrv1.MutatingWebhookConfiguration{}
+	whConfig.Name = csvWebhook.Name
+
+	res, err := controllerutil.CreateOrUpdate(ctx, cli, whConfig, func() error {
+
+		// openshift fills in the ca on finding this annotation
+		whConfig.Annotations = map[string]string{
+			"service.beta.openshift.io/inject-cabundle": "true",
+		}
+
```
nb-ohad, line 143: Let not overwrite any other annotation introduced on the web hook
```suggestion
		if whConfig.Annotations == nil {
			whConfig.Annotations = map[string]string{}
		}
		whConfig.Annotations["service.beta.openshift.io/inject-cabundle"] = "true'
```

controllers/webhook.go
```diff
@@ -0,0 +1,189 @@
...
+		var caBundle []byte
+		if len(whConfig.Webhooks) == 0 {
+			whConfig.Webhooks = make([]admrv1.MutatingWebhook, 1)
+		} else {
+			// do not mutate CA bundle that was injected by openshift
+			caBundle = whConfig.Webhooks[0].ClientConfig.CABundle
+		}
+
+		// webhook desired state
+		var wh *admrv1.MutatingWebhook = &whConfig.Webhooks[0]
```
nb-ohad, line 153:
```suggestion
		admrv1.MutatingWebhook := &whConfig.Webhooks[0]
```

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type ClusterServiceVersionDefaulter struct {
+	client.Client
+
+	Decoder           admission.Decoder
+	OperatorNamespace string
+
+	ctx context.Context
```
nb-ohad, line 43: Similar to one of the previous comments (one of the first comments), there should not be any request related state set on a struct that is shared between multiple requests

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	log logr.Logger
+}
+
+// +kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;patch
+
+func (r *ClusterServiceVersionDefaulter) Handle(ctx context.Context, req admission.Request) admission.Response {
+
+	r.ctx = ctx
+	r.log = log.FromContext(ctx)
+	r.log.Info("request received for csv review")
```
nb-ohad, line 53:
```suggestion
	logger = log.FromContext(ctx)
	logger.Info("request received for csv review")
```

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	marshaledCsv, err := json.Marshal(csv)
+	if err != nil {
+		r.log.Error(err, "failed marshaling csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed marshaling csv: %v", err))
+	}
+
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledCsv)
+}
+
+func (r *ClusterServiceVersionDefaulter) getOldCsvReplicas(csv *opv1a1.ClusterServiceVersion) (int32, error) {
```
nb-ohad, line 83: Old has a different meaning (it reflect on age not time), let's use previous instead
```suggestion
func (r *ClusterServiceVersionDefaulter) getPreviousCsvReplicas(csv *opv1a1.ClusterServiceVersion) (int32, error) {
```

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	if ok := r.isOdfManagedCsv(csv); !ok {
+		r.log.Info("ignoring requested csv as it is not relevant")
+		return admission.Allowed("csv is not relevant to the odf")
+	}
+
+	replicas, err := r.getOldCsvReplicas(csv)
+	if err != nil {
+		r.log.Error(err, "failed getting replicas from the csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed getting replicas from the csv: %v", err))
+	}
```
nb-ohad, line 70: Why us that separate from `mutateCsv`?

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func (r *ClusterServiceVersionDefaulter) getOldCsvReplicas(csv *opv1a1.ClusterServiceVersion) (int32, error) {
+
+	if csv.Spec.Replaces == "" {
+		r.log.Info("csv.Spec.Replaces is not populated")
+		return 0, nil
+	}
+
+	oldCsv := &opv1a1.ClusterServiceVersion{}
```
nb-ohad, line 90: Same reasoning as last comment
```suggestion
	prevCsv := &opv1a1.ClusterServiceVersion{}
```

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	key := client.ObjectKey{Name: csv.Spec.Replaces, Namespace: csv.Namespace}
+
+	if err := r.Client.Get(r.ctx, key, oldCsv); err != nil {
+		if errors.IsNotFound(err) {
+			// new install where an older csv does not exists
+			return 0, nil
+		}
+		r.log.Error(err, "failed getting older csv")
+		return 0, err
+	}
```
nb-ohad, line 100: Prefer single nesting for handling of a single client result
```suggestion
	if err := r.Client.Get(r.ctx, key, oldCsv); client.IgnoreNotFound(err) != nil {
		r.log.Error(err, "failed getting older csv")
		return 0, err
	} else err != nil {
		// new install where an older csv does not exists
		return 0, nil
	}
```

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+
+	for _, deployment := range oldCsv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		if *deployment.Spec.Replicas > 0 {
+			// upgrade case where an older csv is found with replica 1
+			return 1, nil
```
nb-ohad, line 105: There is a real issue with the logic of this function.
Different deployments inside the same CSV might have different values for replicas. Operator's usually use a single replica, but not all deployment inside of a CSV are operator deployments. 

This function returns a single value, which cannot express the exact state of the previous CSV. 
To fix this I would say we should return a bool and not mutate if the return value is true - which means that a previous CSV had at least one deployment with replica count bigger then 0

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+	return 0, nil
+}
+
+func (r *ClusterServiceVersionDefaulter) mutateCsv(csv *opv1a1.ClusterServiceVersion, replicas *int32) {
+
+	r.log.Info("mutating requested csv")
+	for i := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec.Replicas = replicas
+	}
+}
```
nb-ohad, line 119: Why to we need a function for this code snippet that is only called from a single location?

webhook/csv.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if csv.Namespace != r.OperatorNamespace {
+		return false
+	}
+
+	for _, pkgName := range controllers.PkgNames {
+		if strings.HasPrefix(csv.Name, pkgName) {
+			return true
+		}
+	}
```
nb-ohad, line 131: You cannot run against a hard coded list, even if it a calculated list. You should run on the list loaded from the CM which is the `ResourceMappingList` list because it is guaranteed to update at run time

nb-ohad (changes requested):

controllers/operatorscaler_controller.go
```diff
@@ -93,6 +93,13 @@ var (
...
+		for _, resMap := range ResourceMappingList {
+			pkgNames = append(pkgNames, resMap.PkgNames...)
+		}
+		return
+	}()
```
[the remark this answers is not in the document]
iamniting: It does have a return value `pkgNames` which is defined in the func signature.

controllers/storagesystem_controller.go
iamniting: This controller anyway will be removed.

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
iamniting: Anyway, I removed this and changed to the func.

controllers/storagesystem_controller.go
iamniting: anyway, i removed it.

controllers/storagesystem_controller.go
iamniting: This is removed.

controllers/storagesystem_controller.go
iamniting: Done

controllers/storagesystem_controller.go
iamniting: removed it.

controllers/subscription_controller.go
iamniting: removed it.

controllers/subscription_controller.go
iamniting: removed it.

controllers/subscription_controller.go
iamniting: I removed this and passing the constant name.

controllers/webhook.go
iamniting: Done

controllers/webhook.go
iamniting: I removed these constants, and everything is now inside the template.

controllers/webhook.go
iamniting: It is better call one func for the webhook from the main recociler which fulfills the all requirements of webhook.

controllers/webhook.go
iamniting: I removed the constants.

controllers/webhook.go
iamniting: No i would like to have it as a whole, I removed the constants.

controllers/webhook.go
iamniting: constant is removed.

controllers/webhook.go
iamniting: removed the constant.

controllers/webhook.go
iamniting: Done

controllers/webhook.go
iamniting: Done

webhook/csv.go
iamniting: Done

webhook/csv.go
iamniting: Done

webhook/csv.go
iamniting: changed the name to the `isPreviousCsvRunningDeployments`

webhook/csv.go
iamniting: Done

webhook/csv.go
iamniting: Done

webhook/csv.go
iamniting: Done

controllers/webhook.go
iamniting: Done

nb-ohad (changes requested):

config/manager/webhook_service.yaml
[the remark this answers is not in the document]
iamniting: We need the secret to be present before running the odf-operator, so that pod can mount it.

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type ClusterServiceVersionDefaulter struct {
```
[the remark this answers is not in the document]
iamniting: `defualter` is used for the mutating webhook because it defaults the crd.

https://github.com/search?q=repo%3Akubernetes-sigs%2Fkubebuilder%20Defaulter&type=code

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+		return admission.Allowed("csv is not relevant to the odf")
+	}
+
+	up, err := r.isPreviousCsvRunningDeployments(ctx, logger, csv)
+	if err != nil {
+		logger.Error(err, "failed getting replicas from the csv")
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed getting replicas from the csv: %v", err))
+	}
+
+	r.mutateCsv(logger, csv, up)
```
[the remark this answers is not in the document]
iamniting: update is too vague IMO. `mutate` or we can use `default` which makes more sense in the webhook. even `default ` is created by scafolding.

controllers/mapping.go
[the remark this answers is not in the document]
iamniting: changed to `crd_mapping.go`

controllers/storagesystem_controller.go
```diff
@@ -44,6 +44,10 @@ const (
 	storageSystemFinalizer = "storagesystem.odf.openshift.io"
 )
 
+var (
+	IsCsvWebhookCreated = false
+)
```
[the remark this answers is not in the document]
iamniting: I removed it and doing a get call.

controllers/storagesystem_controller.go
```diff
@@ -70,6 +74,10 @@ type StorageSystemReconciler struct {
 func (r *StorageSystemReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logger := r.Log.WithValues("instance", req.NamespacedName)
 
+	if !IsCsvWebhookCreated {
```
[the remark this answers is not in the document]
iamniting: I removed it and doing a get call.

controllers/subscription_controller.go
```diff
@@ -75,12 +78,19 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	err = r.setOperatorCondition(logger, r.OperatorNamespace)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	err = r.ensureSubscriptions(logger, req.NamespacedName.Namespace)
+	err = reconcileCsvWebhook(ctx, r.Client, logger, r.OperatorNamespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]

controllers/subscription_controller.go
```diff
@@ -75,12 +78,19 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	IsCsvWebhookCreated = true
```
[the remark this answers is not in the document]
iamniting: removed it.

controllers/subscription_controller.go
```diff
@@ -290,12 +300,19 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	)
 
+	enqueueOdfSub := handler.EnqueueRequestsFromMapFunc(
+		func(ctx context.Context, obj client.Object) []reconcile.Request {
+			return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: "odf-operator", Namespace: r.OperatorNamespace}}}
```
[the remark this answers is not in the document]

controllers/webhook.go
```diff
@@ -0,0 +1,192 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func reconcileCsvMutatingWebhookService(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) error {
+
+	svc := &corev1.Service{}
+	svc.Name = csvWebhookService.ObjectMeta.Name
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+	logger := log.FromContext(ctx)
+	logger.Info("request received for csv review")
+
+	csv := &opv1a1.ClusterServiceVersion{}
+	if err := r.Decoder.Decode(req, csv); err != nil {
+		logger.Error(err, "failed decoding admission review as csv")
+		return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed decoding admission review as csv: %v", err))
+	}
+
+	if ok := r.isOdfManagedCsv(csv); !ok {
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+		logger.Info("ignoring requested csv as it is not relevant")
+		return admission.Allowed("csv is not relevant to the odf")
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	up, err := r.isPreviousCsvRunningDeployments(ctx, logger, csv)
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		logger.Error(err, "failed getting replicas from the csv")
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed getting replicas from the csv: %v", err))
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	prevCsv := &opv1a1.ClusterServiceVersion{}
+	key := client.ObjectKey{Name: csv.Spec.Replaces, Namespace: csv.Namespace}
+
+	if err := r.Client.Get(ctx, key, prevCsv); errors.IsNotFound(err) {
+		// new install where an older csv does not exists
+		return false, nil
+	} else if err != nil {
+		logger.Error(err, "failed getting older csv")
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+		return false, err
+	}
+
+	for _, deployment := range prevCsv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
```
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+	logger.Info("mutating requested csv")
+
+	replicas := ptr.To(int32(0))
+	if up {
+		replicas = ptr.To(int32(1))
+	}
+
+	for i := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec.Replicas = replicas
+	}
```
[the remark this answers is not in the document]
iamniting: I made the changes. Now i am not changing the csv if any deployment was up in the prev csv.

webhook/csv.go
```diff
@@ -0,0 +1,143 @@
...
+		return false
+	}
+
+	for _, pkgName := range controllers.GetPkgNames() {
+		if strings.HasPrefix(csv.Name, pkgName) {
+			return true
+		}
+	}
+
+	return false
```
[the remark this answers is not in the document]
iamniting: I don't understand what is the issue with the above approach.

config/manager/webhook_service.yaml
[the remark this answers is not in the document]
nb-ohad: I am not sure I follow. Can you please elaborate?

webhook/csv.go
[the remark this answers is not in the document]
nb-ohad: But you are not defaulting. You are deciding on a custom value based on logic. 
Even if I accept that "Defaulter" is a word (as in the example you gave, it definitely does not apply here. 

Why not something like `ClusterServiceVersionDeploymentScaler`?

webhook/csv.go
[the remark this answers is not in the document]
nb-ohad: But this has nothing with default values. If you are against `update' I would say let's be more specific:

`scaleDownCsvDeployments(logger, csv)` 

And drop the up value completely as the logic should be

```suggestion
	if !running {
		r.scaleDownCsvDeployments(logger, csv)
	}
```

webhook/csv.go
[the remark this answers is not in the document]
nb-ohad: you should not even call the function if there is no need to touch the deployments. see this comment: https://github.com/red-hat-storage/odf-operator/pull/516#discussion_r2001062074

webhook/csv.go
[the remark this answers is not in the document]
nb-ohad: If the package you are looking for is the first package, why do we need to create an in memory slice of package names? It has both memory implications and execution implications

config/manager/webhook_service.yaml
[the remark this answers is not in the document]
iamniting: This service has an annotation that will create a secret. The secret is mounted in the odf-operator deployment in order to work with the webhook.

webhook/csv.go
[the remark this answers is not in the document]
iamniting: But it is `defaulting` based on some logic. Otherwise people can also use the kubebuilder tags for the default values.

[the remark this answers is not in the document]
iamniting: Changed it to `ClusterServiceVersionDeploymentScaler` as discusses over a call.

webhook/csv.go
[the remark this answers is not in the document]
iamniting: Done a bit differently.

webhook/csv.go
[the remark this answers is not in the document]
iamniting: Done

webhook/csv.go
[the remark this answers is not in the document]
iamniting: Done

config/manager/webhook_service.yaml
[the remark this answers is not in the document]
nb-ohad: Discussed offline and where we agreed that there is a risk in having 2 locations who act as a source of truth, but at the same time we don't want to block this PR. We will convert this discussion into a an issue and resolve the thread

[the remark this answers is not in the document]
nb-ohad: Issue can be found here: https://github.com/red-hat-storage/odf-operator/issues/560

PR #520, opened 2025-01-28
godeps: fix an error 'unrecognized import path "go.opencensus.io"'

iamniting:
github.com/census-instrumentation/opencensus-go repository was archived on the Jul 31, 2023 and url go.opencensus.io holding these libraries is not available anymore. Though it is still available on the proxy server.

But we are using GOPROXY="direct,https://proxy.golang.org" which tries to fetch it from the go.opencensus.io and throws an error. To fix this we are now starting pulling it directly from the github archived repo.

once we start using GOPROXY="https://proxy.golang.org,direct" we can revert this change as well.

while reverting 6cb13e914938e7c8d2832fde0b930cebcb84a98e we also need to revert this. Related PR (#479)

Ref: 
https://github.com/googleapis/google-cloud-go/issues/11481 
https://github.com/kubernetes-sigs/kubetest2/pull/287

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -113,6 +113,8 @@ require (
 
 replace github.com/portworx/sched-ops => github.com/portworx/sched-ops v0.20.4-openstorage-rc3 // required by rook v1.7
 
+replace go.opencensus.io => github.com/census-instrumentation/opencensus-go v0.23.0
+
 exclude (
 	// This tag doesn't exist, but is imported by github.com/portworx/sched-ops.
 	github.com/kubernetes-incubator/external-storage v0.20.4-openstorage-rc2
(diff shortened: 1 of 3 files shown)


Nikhil-Ladha:
> once we start using GOPROXY="https://proxy.golang.org,direct/" we can revert this change as well.

When are we planning to do so, and what's the impact if we do so now?

iamniting:
> > once we start using GOPROXY="https://proxy.golang.org,direct/" we can revert this change as well.
> 
> When are we planning to do so, and what's the impact if we do it now?

We will do that once we update the IBM odf operator APIs as we are pulling them directly from github instead of a proxy server.

PR #521, opened 2025-01-29
dockerfile: fix casing warning in Dockerfile

iamniting:
Docker requires consistent casing for keywords.
If a keyword is in uppercase, all keywords should be in uppercase. 
If a keyword is in lowercase, all keywords should be in lowercase.

The following warning appears while building images:

FromAsCasing: 'as' and 'FROM' keywords' casing do not match


Nikhil-Ladha (approved):
/lgtm

PR #529, opened 2025-02-13
create a operator scaler controller to add dynamic watches on CRs and scale operators

iamniting:

### DIFF
--- a/controllers/operatorscaler_controller.go
+++ b/controllers/operatorscaler_controller.go
@@ -1,0 +1,330 @@
+/*
+Copyright 2025 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"go.uber.org/multierr"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// csvLabelKey is a label key to identify CSV with pkg name and namespace
+	CsvLabelKey = "operators.coreos.com/%s.%s"
+)
+
+type ResourceMappingRecord struct {
+	CrdName    string
+	ApiVersion string
+	Kind       string
+	PkgNames   []string
+}
+
+var (
+	ResourceMappingList = []ResourceMappingRecord{
+		{
+			CrdName:    "storageclusters.ocs.openshift.io",
+			ApiVersion: "ocs.openshift.io/v1",
+			Kind:       "StorageCluster",
+			PkgNames:   []string{OcsSubscriptionPackage},
+		},
+		{
+			CrdName:    "cephclusters.ceph.rook.io",
+			ApiVersion: "ceph.rook.io/v1",
+			Kind:       "CephCluster",
+			PkgNames:   []string{RookSubscriptionPackage, CephCSISubscriptionPackage, CSIAddonsSubscriptionPackage},
+		},
+		{
+			CrdName:    "noobaas.noobaa.io",
+			ApiVersion: "noobaa.io/v1alpha1",
+			Kind:       "NooBaa",
+			PkgNames:   []string{NoobaaSubscriptionPackage},
+		},
+		{
+			CrdName:    "prometheuses.monitoring.coreos.com",
+			ApiVersion: "monitoring.coreos.com/v1",
+			Kind:       "Prometheus",
+			PkgNames:   []string{PrometheusSubscriptionPackage},
+		},
+		{
+			CrdName:    "alertmanagers.monitoring.coreos.com",
+			ApiVersion: "monitoring.coreos.com/v1",
+			Kind:       "Alertmanager",
+			PkgNames:   []string{PrometheusSubscriptionPackage},
+		},
+		{
+			CrdName:    "flashsystemclusters.odf.ibm.com",
+			ApiVersion: "odf.ibm.com/v1alpha1",
+			Kind:       "FlashSystemCluster",
+			PkgNames:   []string{IbmSubscriptionPackage},
+		},
+	}
+
+	createOnlyPredicate = predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return true
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+)
+
+type OperatorScalerReconciler struct {
+	client.Client
+
+	Scheme            *runtime.Scheme
+	OperatorNamespace string
+
+	ctx               context.Context
+	log               logr.Logger
+	cache             cache.Cache
+	controller        controller.Controller
+	kindsBeingWatched map[string]bool
+}
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=operators.coreos.com,resources=clusterserviceversions,verbs=get;list;update
+
+//+kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=get;list;watch
+//+kubebuilder:rbac:groups=ceph.rook.io,resources=cephclusters,verbs=get;list;watch
+//+kubebuilder:rbac:groups=noobaa.io,resources=noobaas,verbs=get;list;watch
+//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheuses,verbs=get;list;watch
+//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=alertmanagers,verbs=get;list;watch
+//+kubebuilder:rbac:groups=odf.ibm.com,resources=flashsystemclusters,verbs=get;list;watch
+
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
+func (r *OperatorScalerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	r.ctx = ctx
+	r.log = log.FromContext(ctx)
+
+	r.log.Info("starting reconcile")
+
+	if err := r.isOdfDependenciesCsvReady(); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.reconcileOperators(); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.reconcileDynamicWatchers(); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.log.Info("reconcile completed successfully")
+	return ctrl.Result{}, nil
+}
+
+func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady() error {
+	r.log.Info("entering isOdfDependenciesCsvReady")
+
+	odfDepsCsv := &opv1a1.ClusterServiceVersion{}
+
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: OdfDepsSubscriptionStartingCSV, Namespace: r.OperatorNamespace}, odfDepsCsv)
+	if err != nil {
+		r.log.Error(err, "failed getting odf-deps csv", "csvName", OdfDepsSubscriptionStartingCSV)
+		return err
+	}
+
+	if odfDepsCsv.Status.Phase != opv1a1.CSVPhaseSucceeded {
+		err = fmt.Errorf("csv %s is not in succeeded state", OdfDepsSubscriptionStartingCSV)
+		r.log.Error(err, "waiting for csv to be in succeeded state")
+		return err
+	}
+
+	r.log.Info("successfully completed isOdfDependenciesCsvReady")
+	return nil
+}
+
+func (r *OperatorScalerReconciler) reconcileOperators() error {
+	r.log.Info("entering reconcileOperators")
+
+	var returnErr error
+
+	for i := range ResourceMappingList {
+		resourceMapping := &ResourceMappingList[i]
+
+		crList := &metav1.PartialObjectMetadataList{}
+		crList.TypeMeta.APIVersion = resourceMapping.ApiVersion
+		crList.TypeMeta.Kind = resourceMapping.Kind
+
+		if err := r.Client.List(r.ctx, crList, client.Limit(1)); err != nil {
+			if !meta.IsNoMatchError(err) {
+				msg := fmt.Sprintf("failed listing %s", resourceMapping.Kind)
+				r.log.Error(err, msg)
+				multierr.AppendInto(&returnErr, err)
+			}
+
+		} else if len(crList.Items) > 0 {
+
+			for _, pkgName := range resourceMapping.PkgNames {
+				key := fmt.Sprintf(CsvLabelKey, pkgName, r.OperatorNamespace)
+
+				csvList := &opv1a1.ClusterServiceVersionList{}
+				err = r.Client.List(
+					r.ctx, csvList,
+					client.InNamespace(r.OperatorNamespace),
+					client.MatchingLabels(map[string]string{key: ""}),
+				)
+				if err != nil {
+					r.log.Error(err, "failed listing csvs with label", "label", key)
+					multierr.AppendInto(&returnErr, err)
+				} else {
+					for i := range csvList.Items {
+						if err = r.updateCsvDeplymentsReplicas(&csvList.Items[i]); err != nil {
+							r.log.Error(err, "failed updating csvs replica")
+							multierr.AppendInto(&returnErr, err)
+						}
+					}
+				}
+			}
+		}
+	}
+
+	r.log.Info("successfully completed reconcileOperators")
+	return returnErr
+}
+
+func (r *OperatorScalerReconciler) updateCsvDeplymentsReplicas(csv *opv1a1.ClusterServiceVersion) error {
+
+	var updateRequired bool
+	for i := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+		deploymentSpec := &csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec
+		if *deploymentSpec.Replicas < 1 {
+			deploymentSpec.Replicas = ptr.To(int32(1))
+			updateRequired = true
+		}
+	}
+
+	if updateRequired {
+		if err := r.Client.Update(r.ctx, csv); err != nil {
+			r.log.Error(err, "failed updating csv replica", "csvName", csv.Name)
+			return err
+		}
+		r.log.Info("csv updated successfully", "csvName", csv.Name)
+	}
+
+	return nil
+}
+
+func (r *OperatorScalerReconciler) reconcileDynamicWatchers() error {
+	r.log.Info("entering reconcileDynamicWatchers")
+
+	for i := range ResourceMappingList {
+		resourceMapping := &ResourceMappingList[i]
+
+		if !r.kindsBeingWatched[resourceMapping.Kind] {
+
+			crd := &extv1.CustomResourceDefinition{}
+			crd.Name = resourceMapping.CrdName
+			if err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(crd), crd); client.IgnoreNotFound(err) != nil {
+				r.log.Error(err, "failed getting crd", "crdName", resourceMapping.CrdName)
+				return err
+			} else if err == nil {
+				r.log.Info("adding dynamic watch", "kind", resourceMapping.Kind)
+
+				err := r.controller.Watch(
+					source.Kind(
+						r.cache,
+						client.Object(
+							&metav1.PartialObjectMetadata{
+								TypeMeta: metav1.TypeMeta{
+									APIVersion: resourceMapping.ApiVersion,
+									Kind:       resourceMapping.Kind,
+								},
+							},
+						),
+						&handler.EnqueueRequestForObject{},
+						// Trigger the reconcile for creation events of the object.
+						// This ensures the replicas in the CSV are scaled up based on the presence of Custom Resource (CR).
+						createOnlyPredicate,
+					),
+				)
+				if err != nil {
+					r.log.Error(err, "failed adding dynamic watch", "kind", resourceMapping.Kind)
+					return err
+				}
+
+				r.kindsBeingWatched[resourceMapping.Kind] = true
+			}
+		}
+	}
+
+	r.log.Info("successfully completed reconcileDynamicWatchers")
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *OperatorScalerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	// It will keep track or the crds that we care about
+	crdMap := map[string]bool{}
+	for i := range ResourceMappingList {
+		crdName := ResourceMappingList[i].CrdName
+		crdMap[crdName] = true
+	}
+
+	controller, err := ctrl.NewControllerManagedBy(mgr).
+		Named("operatorScaler").
+		Watches(
+			&extv1.CustomResourceDefinition{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				predicate.NewPredicateFuncs(func(obj client.Object) bool {
+					return crdMap[obj.GetName()]
+				}),
+				// Trigger a reconcile only during the creation of a specific CRD to ensure it runs exactly once for that CRD.
+				// This is required to dynamically add a watch for the corresponding Custom Resource (CR) based on the CRD name.
+				// The reconcile will be triggered with the CRD name as `req.Name`, and the reconciler will set up a watch for the CR using the CRD name.
+				createOnlyPredicate,
+			),
+		).
+		Build(r)
+
+	r.kindsBeingWatched = map[string]bool{}
+	r.controller = controller
+	r.cache = mgr.GetCache()
+
+	return err
+}
--- a/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
@@ -193,6 +193,14 @@ spec:
           verbs:
           - update
         - apiGroups:
+          - ceph.rook.io
+          resources:
+          - cephclusters
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
           - config.openshift.io
           resources:
           - clusterversions
@@ -237,6 +245,23 @@ spec:
           - get
           - list
           - update
+        - apiGroups:
+          - monitoring.coreos.com
+          resources:
+          - alertmanagers
+          - prometheuses
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - noobaa.io
+          resources:
+          - noobaas
+          verbs:
+          - get
+          - list
+          - watch
         - apiGroups:
           - ocs.openshift.io
           resources:
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -51,6 +51,14 @@ rules:
   verbs:
   - update
 - apiGroups:
+  - ceph.rook.io
+  resources:
+  - cephclusters
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - config.openshift.io
   resources:
   - clusterversions
@@ -95,6 +103,23 @@ rules:
   - get
   - list
   - update
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - alertmanagers
+  - prometheuses
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - noobaa.io
+  resources:
+  - noobaas
+  verbs:
+  - get
+  - list
+  - watch
 - apiGroups:
   - ocs.openshift.io
   resources:
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.32.2
 	k8s.io/apimachinery v0.32.2
 	k8s.io/client-go v0.32.2
+	k8s.io/utils v0.0.0-20241210054802-24370beab758
 	sigs.k8s.io/controller-runtime v0.20.2
 )
 
@@ -129,7 +130,6 @@ require (
 	k8s.io/component-base v0.32.2 // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20241212222426-2c72e554b1e7 // indirect
-	k8s.io/utils v0.0.0-20241210054802-24370beab758 // indirect
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.0 // indirect
 	sigs.k8s.io/container-object-storage-interface-api v0.1.0 // indirect
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
--- a/main.go
+++ b/main.go
@@ -170,6 +170,15 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "ClusterVersion")
 		os.Exit(1)
 	}
+
+	if err = (&controllers.OperatorScalerReconciler{
+		Client:            mgr.GetClient(),
+		Scheme:            mgr.GetScheme(),
+		OperatorNamespace: operatorNamespace,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "OperatorScaler")
+		os.Exit(1)
+	}
 	//+kubebuilder:scaffold:builder
 
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {


controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+			r.logger.Error(err, "failed to list objects")
+			return err
+		}
+
+		if len(objects.Items) == 0 {
+			replicas = 0
+		}
+
+		for _, csvName := range csvNames {
```
leelavg, line 185: maybe it's better to run this in parallel or let the loop go over all CSVs trying to update CSV as a failure in one stops everything else now, why do want to put a break if all operators can be run in parallel in the cluster?

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	r.logger.Info("ending reconcile")
+	return ctrl.Result{}, nil
+}
+
+func (r *ScalerReconciler) scaleOperators() error {
+
+	for kind, csvNames := range kindToCsvsMapping {
+
+		var replicas int32 = 1
+		objects := &metav1.PartialObjectMetadataList{}
```
leelavg, line 169: possible to give a more concrete name here? these are CRs, isn't it? crList?

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+		objects.TypeMeta = kind.TypeMeta
```
leelavg, line 170: does this work 🤔? isn't List and individual resource distinct? this line here seems like assigning Pod typeMeta to PodList typeMeta.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+		err := r.Client.List(r.ctx, objects)
+		if err != nil {
+			if meta.IsNoMatchError(err) {
+				continue
+			}
```
leelavg, line 176: is `client.IgnoreNotFound(err) != nil` different than `IsNoMatchError`?

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+					UpdateFunc: func(e event.UpdateEvent) bool {
+						return false
+					},
+					GenericFunc: func(e event.GenericEvent) bool {
+						return false
+					},
+				},
+				predicate.GenerationChangedPredicate{},
+			),
+		).
```
leelavg, line 302: i'm just thinking, do we want to add a watch for CSV as we are effectively managing it's replicas?

leelavg (commented):
will review again after some of the comments are addressed.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	csvOcsClientOperator     = "ocs-client-operator" //nolint:unused
+	csvOcsOperator           = "ocs-operator"
+	csvOdfCsiAddonsOperator  = "odf-csi-addons-operator" //nolint:unused
+	csvOdfDependencies       = "odf-dependencies"        //nolint:unused
+	csvOdfOperator           = "odf-operator"            //nolint:unused
+	csvOdfPrometheusOperator = "odf-prometheus-operator" //nolint:unused
+	csvRecipe                = "recipe"                  //nolint:unused
+	csvRookCephOperator      = "rook-ceph-operator"
+
+	csvIbmStorageOdfOperator = "ibm-storage-odf-operator"
```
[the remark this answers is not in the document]
iamniting: yes, i Just named as CSV as we refer to csv mostly.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var csvLabelKey = "operators.coreos.com/%s.%s"
+
+var (
```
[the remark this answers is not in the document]
iamniting: I am doing that to keep different logical variables separate.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+		kindCephCluster:        []string{csvRookCephOperator},
+		kindNoobaa:             []string{csvMcgOperator},
+		kindFlashSystemCluster: []string{csvIbmStorageOdfOperator},
+	}
+)
+
+type ScalerReconciler struct {
+	ctx               context.Context
+	logger            logr.Logger
+	Client            client.Client
```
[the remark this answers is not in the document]
iamniting: Actually i want that all the client APIS and func gets called via the client itself and not directly via the reconciler. It reduce confusions.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	err = r.scaleOperators()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.logger.Info("ending reconcile")
+	return ctrl.Result{}, nil
+}
+
+func (r *ScalerReconciler) scaleOperators() error {
```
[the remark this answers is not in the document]
iamniting: The one i am following now is operator-sdk layout where `Reconcile` is the first one and `SetupWithManager` is the last one.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+		kindStorageCluster:     []string{csvOcsOperator},
+		kindCephCluster:        []string{csvRookCephOperator},
+		kindNoobaa:             []string{csvMcgOperator},
+		kindFlashSystemCluster: []string{csvIbmStorageOdfOperator},
+	}
+)
+
+type ScalerReconciler struct {
+	ctx               context.Context
+	logger            logr.Logger
```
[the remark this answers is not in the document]
iamniting: I believe anything should be fine. Both make sense.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+var (
+	crdToKindMapping = map[string]metav1.PartialObjectMetadata{
+		crdStorageCluster:     *kindStorageCluster,
+		crdCephCluster:        *kindCephCluster,
+		crdNoobaa:             *kindNoobaa,
+		crdFlashSystemCluster: *kindFlashSystemCluster,
+	}
+
+	kindToCsvsMapping = map[*metav1.PartialObjectMetadata][]string{
```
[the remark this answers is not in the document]
iamniting: yes thats correct. Even when we will enhance this we will need more than one csvs to be scaled.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+			if meta.IsNoMatchError(err) {
+				continue
+			}
+			r.logger.Error(err, "failed to list objects")
+			return err
+		}
+
+		if len(objects.Items) == 0 {
+			replicas = 0
+		}
```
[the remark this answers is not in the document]
iamniting: correct

controllers/scaler_controller.go
iamniting: currently i dont want to because if we do that we don't leave any room for user to make the change to the replicas of the CSV also they wont be able to scale down the odf-operator because of the webhook.

controllers/scaler_controller.go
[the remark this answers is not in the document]
leelavg: No, I mean these represent pkg and should be named accordingly.

controllers/scaler_controller.go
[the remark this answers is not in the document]
leelavg: That isn't helping anything I believe, a single var w/ a comment around each segment would inform the intent more appropriately.

In current state, reviewers (future & current) should ensure ntg in b/n should be written.

controllers/scaler_controller.go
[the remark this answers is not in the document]
leelavg: I don't think so, it's contrary to your statement ie, not embedding w/o any upside is different to how reconcilers are scaffolded.

controllers/scaler_controller.go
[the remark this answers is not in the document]
leelavg: Yes, they are first and last as there is ntg in b/n exported, so all other methods off of reconciler and helper should come after exported functions then?

controllers/scaler_controller.go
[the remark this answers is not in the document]
iamniting: renamed them.

controllers/scaler_controller.go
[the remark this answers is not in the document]
iamniting: created under one block.

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	crdCephCluster        = "cephclusters.ceph.rook.io"
+	crdNoobaa             = "noobaas.noobaa.io"
+	crdFlashSystemCluster = "flashsystemclusters.odf.ibm.com"
+
+	crdList = []string{
+		crdStorageCluster,
+		crdCephCluster,
+		crdNoobaa,
+		crdFlashSystemCluster,
+	}
```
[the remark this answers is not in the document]
iamniting: done

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	csvOdfOperator           = "odf-operator"            //nolint:unused
+	csvOdfPrometheusOperator = "odf-prometheus-operator" //nolint:unused
+	csvRecipe                = "recipe"                  //nolint:unused
+	csvRookCephOperator      = "rook-ceph-operator"
+
+	csvIbmStorageOdfOperator = "ibm-storage-odf-operator"
+)
+
+var (
+	crdToKindMapping = map[string]metav1.PartialObjectMetadata{
```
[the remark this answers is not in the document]
iamniting: made changes as per 2

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"slices"
+
+	"github.com/go-logr/logr"
+	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
```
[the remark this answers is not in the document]
iamniting: changed it to `oprv1a1`

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+func (r *ScalerReconciler) addWatches(req ctrl.Request) error {
+
+	if !slices.Contains(crdList, req.Name) {
+		return nil
+	}
+
+	r.logger.Info("adding dynamic watch", "object", req.Name)
+
+	kind := crdToKindMapping[req.Name]
+	err := r.addDynamicWatch(&kind)
```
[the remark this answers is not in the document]
iamniting: done

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	return nil
+}
+
+func (r *ScalerReconciler) addDynamicWatch(kind *metav1.PartialObjectMetadata) error {
+	return r.controller.Watch(
+		source.Kind(
+			r.mgr.GetCache(),
+			client.Object(kind),
```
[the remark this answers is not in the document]
iamniting: done

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+						return false
+					},
+				},
+				predicate.GenerationChangedPredicate{},
+			),
+		).
+		Build(r)
+
+	r.controller = controller
+	r.mgr = mgr
```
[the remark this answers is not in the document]
iamniting: storing cache only

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+	kindToCsvsMapping = map[*metav1.PartialObjectMetadata][]string{
+		kindStorageCluster:     []string{csvOcsOperator},
+		kindCephCluster:        []string{csvRookCephOperator},
+		kindNoobaa:             []string{csvMcgOperator},
+		kindFlashSystemCluster: []string{csvIbmStorageOdfOperator},
+	}
+)
+
+type ScalerReconciler struct {
```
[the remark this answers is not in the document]
iamniting: changed it to operator scaler reconciler

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+					r.logger.Info("csv updated successfully", "csvName", csvObj.Name, "replicas", replicas)
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *ScalerReconciler) addWatches(req ctrl.Request) error {
```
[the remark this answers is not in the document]
iamniting: using the name only now

controllers/scaler_controller.go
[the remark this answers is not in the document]
iamniting: changed it to log

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
+/*
+Copyright 2021 Red Hat OpenShift Data Foundation.
```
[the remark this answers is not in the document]
iamniting: changed it to 2025

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+		*kindNoobaa,
+		*kindFlashSystemCluster,
+	}
+)
+
+var (
+	crdStorageCluster     = "storageclusters.ocs.openshift.io"
+	crdCephCluster        = "cephclusters.ceph.rook.io"
+	crdNoobaa             = "noobaas.noobaa.io"
+	crdFlashSystemCluster = "flashsystemclusters.odf.ibm.com"
```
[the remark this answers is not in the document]
iamniting: created as const

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+	crdList = []string{
+		crdStorageCluster,
+		crdCephCluster,
+		crdNoobaa,
+		crdFlashSystemCluster,
+	}
+)
+
+var (
```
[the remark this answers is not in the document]
iamniting: done

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+					LabelSelector: labels.SelectorFromSet(map[string]string{key: ""}),
+					Namespace:     r.OperatorNamespace,
+				},
+			)
+			if err != nil {
+				r.logger.Error(err, "failed to list csvs of label", "label", key)
+				return err
+			}
+
+			for _, csvObj := range csvList.Items {
```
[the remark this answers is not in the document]
iamniting: done

controllers/scaler_controller.go
```diff
@@ -0,0 +1,309 @@
...
+		for _, csvName := range csvNames {
+			key := fmt.Sprintf(csvLabelKey, csvName, r.OperatorNamespace)
+
+			csvList := &operatorv1alpha1.ClusterServiceVersionList{}
+			err = r.Client.List(
+				r.ctx, csvList,
+				&client.ListOptions{
+					LabelSelector: labels.SelectorFromSet(map[string]string{key: ""}),
+					Namespace:     r.OperatorNamespace,
+				},
```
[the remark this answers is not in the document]
iamniting: done

controllers/scaler_controller.go
iamniting: changed it to crList

controllers/scaler_controller.go
iamniting: done

controllers/scaler_controller.go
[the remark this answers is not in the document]
iamniting: changed it.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	return returnErr
+}
+
+func (r *OperatorScalerReconciler) addWatches(crdName string) error {
+
+	if !slices.Contains(crdNameList, crdName) {
+		return nil
+	}
```
nb-ohad, line 304: Why do we need this check? we already set up a predicate that makes sure we are filtering out these

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	r.log = log.FromContext(ctx)
+
+	r.log.Info("starting reconcile")
+
+	err := r.isOdfDependenciesCsvReady()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	err = r.addWatches(req.Name)
```
nb-ohad, line 187: This is a named controller which mean you do not know the type of the object on the request. It might be a CRD or a CR and controllers should not behave differently for different kinds on the request. This leaves us with 2 options:
1. separate into 2 different controllers - I don't think we should do that
2. Target all CRDs in addWatches and don't base it on the request - This would be my preference

nb-ohad, line 187: Can we use reconcilePrefix for the name of the reconcile functions?
```suggestion
	err = r.reconcileDynamicWatchers(req.Name)
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	err = r.scaleOperators()
```
nb-ohad, line 192: Can we use reconcilePrefix for the name of the reconcile functions?
```suggestion
	err = r.reconcileOperators()
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+		source.Kind(
+			r.cache,
+			kind,
+			&handler.EnqueueRequestForObject{},
+			// Trigger the reconcile for creation events of the object.
+			// This ensures the replicas in the CSV are scaled up based on the presence of Custom Resource (CR).
+			createPredicate,
+			predicate.GenerationChangedPredicate{},
+		),
+	)
```
nb-ohad, line 335: This is only used once. Do we really need a function for this?

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+	kind := crdNameToKindMapping[crdName]
+	err := r.addDynamicWatch(kind)
+	if err != nil {
+		r.log.Error(err, "failed to add dynamic watch", "kind", crdName)
+		return err
+	}
+
+	r.watching[crdName] = true
+
```
nb-ohad, line 320: We should be running on all CRDs

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+		crdNameFlashSystemCluster: kindFlashSystemCluster,
+	}
+
+	kindToPkgNamesMapping = map[*metav1.PartialObjectMetadata][]string{
+		kindStorageCluster:     {pkgNameOcsOperator},
+		kindCephCluster:        {pkgNameRookCephOperator, pkgNameCephcsiOperator, pkgNameOdfCsiAddonsOperator},
+		kindNoobaa:             {pkgNameMcgOperator},
+		kindPrometheus:         {pkgNameOdfPrometheusOperator},
+		kindAlertmanager:       {pkgNameOdfPrometheusOperator},
+		kindFlashSystemCluster: {pkgNameIbmStorageOdfOperator},
```
nb-ohad, line 134: This does not make sense as a map. The only place you are using it is as part of a for loop. 
An list will be better, or even preferably collapse the info into the previous map values then run on `maps.Values` in the loop

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+			continue
+		}
+
+		err := r.Client.Update(r.ctx, csv)
+		if err != nil {
+			r.log.Error(err, "failed to update csv replica", "csvName", csv.Name)
+			multierr.AppendInto(&returnErr, err)
+			continue
+		}
+		r.log.Info("csv updated successfully", "csvName", csv.Name)
```
nb-ohad, line 294: Continue is basically an unstructured jump. try to avoid it if possible
```suggestion
		if updateRequired {
			err := r.Client.Update(r.ctx, csv)
			if err != nil {
				r.log.Error(err, "failed to update csv replica", "csvName", csv.Name)
				multierr.AppendInto(&returnErr, err)
			} else {
				r.log.Info("csv updated successfully", "csvName", csv.Name)
			}
		}
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+		err := r.Client.List(r.ctx, crList)
+		if err != nil {
+			if meta.IsNoMatchError(err) {
+				continue
+			}
+			msg := fmt.Sprintf("failed to list %s", kind.TypeMeta.Kind)
+			r.log.Error(err, msg)
+			multierr.AppendInto(&returnErr, err)
+			continue
```
nb-ohad, line 237: Try to minimize you use of `continue`
```suggestion
			if !meta.IsNoMatchError(err) {
				msg := fmt.Sprintf("failed to list %s", kind.TypeMeta.Kind)
				r.log.Error(err, msg)
				multierr.AppendInto(&returnErr, err)
				continue
			}
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+				multierr.AppendInto(&returnErr, err)
+				continue
+			}
+
+			err = r.updateCsvsReplicas(csvList.Items...)
+			if err != nil {
+				r.log.Error(err, "failed to update csvs replica")
+				multierr.AppendInto(&returnErr, err)
+			}
+		}
```
nb-ohad, line 264:
```suggestion
		if len(crList.Items) > 0 {
			for _, pkgName := range pkgNames {
				key := fmt.Sprintf(csvLabelKey, pkgName, r.OperatorNamespace)

				csvList := &oprv1a1.ClusterServiceVersionList{}
				err = r.Client.List(
					r.ctx, csvList,
					client.InNamespace(r.OperatorNamespace),
					client.MatchingLabels(map[string]string{key: ""}),
				)
				if err != nil {
					r.log.Error(err, "failed to list csvs of label", "label", key)
					multierr.AppendInto(&returnErr, err)
				} else {
					err = r.updateCsvsReplicas(csvList.Items...)
					if err != nil {
						r.log.Error(err, "failed to update csvs replica")
						multierr.AppendInto(&returnErr, err)
					}
				}
			}
		}
```

nb-ohad (changes requested):

controllers/operatorscaler_controller.go
iamniting: This is required to make sure that the request is for the CRD and not for the CR.

controllers/operatorscaler_controller.go
iamniting: I believe what I'm doing is more efficient. Targeting all crds means getting the list of crds, looking for the ones we're interested in, and then adding watches. We will be running unnecessary code with every reconciliation which can be avoided with the current approach.

controllers/operatorscaler_controller.go
iamniting: Why doesn’t it make sense as a map? We need this mapping because we have to scale up these CSVs based on the kind.

controllers/operatorscaler_controller.go
iamniting: Eventually it does the same thing. IMO the first one is easy to read and maintain.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	// CRD Names
+	crdNameStorageCluster     = "storageclusters.ocs.openshift.io"
+	crdNameCephCluster        = "cephclusters.ceph.rook.io"
+	crdNameNoobaa             = "noobaas.noobaa.io"
+	crdNamePrometheus         = "prometheuses.monitoring.coreos.com"
+	crdNameAlertmanager       = "alertmanagers.monitoring.coreos.com"
+	crdNameFlashSystemCluster = "flashsystemclusters.odf.ibm.com"
+
+	// OLM pkg Names
+	pkgNameCephcsiOperator       = "cephcsi-operator"
```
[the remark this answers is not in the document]
iamniting: I am doing the camel casing based on the `-`

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	crdNameList = []string{
+		crdNameStorageCluster,
+		crdNameCephCluster,
+		crdNameNoobaa,
+		crdNamePrometheus,
+		crdNameAlertmanager,
+		crdNameFlashSystemCluster,
+	}
```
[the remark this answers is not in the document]
iamniting: removed the list

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+	crdNameToKindMapping = map[string]*metav1.PartialObjectMetadata{
+		crdNameStorageCluster:     kindStorageCluster,
+		crdNameCephCluster:        kindCephCluster,
+		crdNameNoobaa:             kindNoobaa,
+		crdNamePrometheus:         kindPrometheus,
+		crdNameAlertmanager:       kindAlertmanager,
+		crdNameFlashSystemCluster: kindFlashSystemCluster,
```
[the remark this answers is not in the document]
iamniting: `kind<TypeName>` is used multiple times.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+				// Trigger a reconcile only during the creation of a specific CRD to ensure it runs exactly once for that CRD.
+				// This is required to dynamically add a watch for the corresponding Custom Resource (CR) based on the CRD name.
+				// The reconcile will be triggered with the CRD name as `req.Name`, and the reconciler will set up a watch for the CR using the CRD name.
+				createPredicate,
+				predicate.GenerationChangedPredicate{},
+			),
+		).
+		Build(r)
+
+	r.watching = make(map[string]bool)
```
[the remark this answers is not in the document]
iamniting: We need to initialize it, right?

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+type OperatorScalerReconciler struct {
+	client.Client
+	Scheme            *runtime.Scheme
+	OperatorNamespace string
+	ctx               context.Context
+	log               logr.Logger
+	cache             cache.Cache
+	controller        controller.Controller
+	watching          map[string]bool
```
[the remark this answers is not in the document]
iamniting: above one is easy to read IMO

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+	err = r.addWatches(req.Name)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	err = r.scaleOperators()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
iamniting: done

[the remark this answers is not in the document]
iamniting: I would prefer to add watches before scaling up operators.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady() error {
+
+	odfDepsCsv := &oprv1a1.ClusterServiceVersion{}
+
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: OdfDepsSubscriptionStartingCSV, Namespace: r.OperatorNamespace}, odfDepsCsv)
+	if err != nil {
+		r.log.Error(err, "failed to get odf-deps csv", "csvName", OdfDepsSubscriptionStartingCSV)
```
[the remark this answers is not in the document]
iamniting: Does this really matter, Both are same right?

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+
+	r.log.Info("ending reconcile")
+	return ctrl.Result{}, nil
+}
+
+func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady() error {
+
+	odfDepsCsv := &oprv1a1.ClusterServiceVersion{}
+
+	err := r.Client.Get(r.ctx, types.NamespacedName{Name: OdfDepsSubscriptionStartingCSV, Namespace: r.OperatorNamespace}, odfDepsCsv)
```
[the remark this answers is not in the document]
iamniting: It is populated on build time via the build team.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,364 @@
...
+	if odfDepsCsv.Status.Phase != oprv1a1.CSVPhaseSucceeded {
+		err = fmt.Errorf("csv %s is not in succeeded state", OdfDepsSubscriptionStartingCSV)
+		r.log.Error(err, "waiting for csv to be in succeeded state")
+		return err
+	}
+
+	return nil
+}
+
+func (r *OperatorScalerReconciler) scaleOperators() error {
```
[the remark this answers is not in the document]
iamniting: Done

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: You should not do casing based on the value, that does not make sense. Our rules for splitting words (using a `-`) in operator names does not follow splitting over every word

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: My search of the changes shown that it is not. I might be wrong here. Can you please point a reference to where, outside of the context of using the map, it is being used?

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: Separating sections with different accessibility is a common pattern we use everywhere

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: But then you fail if you cannot set watches which is kind of backward

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: Ok, discussed on a call. We don't want to expose this field outside of the package so initialization happens here

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
iamniting: changed the order.

controllers/operatorscaler_controller.go
iamniting: This is removed

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Removed the func as discussed

controllers/operatorscaler_controller.go
iamniting: We are running on all crds now

controllers/operatorscaler_controller.go
iamniting: Changed it to slice as discussed

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+	var returnErr error
+
+	for idx := range csvList {
+		csv := &csvList[idx]
+		var updateRequired bool
+		for i := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+			if *csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec.Replicas != int32(1) {
+				updateRequired = true
+			}
+			csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec.Replicas = ptr.To(int32(1))
```
nb-ohad, line 242: 1. Lets use a ref to the spec to make this more readable
2. use less than instead of not equal
```suggestion
			spec =  &csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec
			if *spec.Replicas < int32(1) {
				spec.Replicas = ptr.To(int32(1))
				updateRequired = true
			}
			
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+		}
+
+		if updateRequired {
+			err := r.Client.Update(r.ctx, csv)
+			if err != nil {
```
nb-ohad, line 247:
```suggestion
			if err := r.Client.Update(r.ctx, csv); err != nil {
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return returnErr
+}
+
+func (r *OperatorScalerReconciler) updateCsvsReplicas(csvList ...oprv1a1.ClusterServiceVersion) error {
```
nb-ohad, line 231: When it comes to code separation, it will be make more sense to have a func that handles a single CSV and update the deploment. That function then can be invoked multiple times (in a for loop) on the calling site
```suggestion
func (r *OperatorScalerReconciler) updateCsvDeplymentsReplicas(csv oprv1a1.ClusterServiceVersion) error {
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+			}
+		}
+	}
+
+	return returnErr
+}
+
+func (r *OperatorScalerReconciler) reconcileDynamicWatchers(crdName string) error {
+
+	for _, metaData := range MetaDataList {
```
nb-ohad, line 261: Prefer taking a reference to creating a local copy
```suggestion
	for i := range MetaDataList {
	    kindMapping := &MetaDataList[i]
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+		if !r.watching[metaData.CrdName] {
+
+			crd := &extv1.CustomResourceDefinition{}
+			err := r.Client.Get(r.ctx, types.NamespacedName{Name: metaData.CrdName}, crd)
+			if err != nil {
```
nb-ohad, line 266: Prefer locality of error variables
```suggestion
			if err := r.Client.Get(r.ctx, types.NamespacedName{Name: metaData.CrdName}, crd); err != nil {
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+				if !errors.IsNotFound(err) {
+					r.log.Error(err, "failed getting crd", "crdName", metaData.CrdName)
+					return err
+				}
+			} else {
+				r.log.Info("adding dynamic watch", "kind", crdName)
+
+				watchErr := r.controller.Watch(
```
nb-ohad, line 274: Just call it `err`
```suggestion
				err := r.controller.Watch(
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	// csvLabelKey is a label key to identify CSV with pkg name and namespace
+	CsvLabelKey = "operators.coreos.com/%s.%s"
+)
+
+type MetaData struct {
```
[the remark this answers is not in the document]
iamniting: I made this public because we will need this info in the webhook.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+func (r *OperatorScalerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	controller, err := ctrl.NewControllerManagedBy(mgr).
+		Named("operatorScaler").
+		Watches(
+			&extv1.CustomResourceDefinition{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				predicate.NewPredicateFuncs(func(obj client.Object) bool {
+					return slices.ContainsFunc(MetaDataList, func(metaData MetaData) bool {
```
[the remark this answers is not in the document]
iamniting: I am doing this for a reason, so that I don't change the usage of watching. It should be watched for what it was created for.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+
+type MetaData struct {
+	CrdName    string
+	ApiVersion string
+	Kind       string
+	PkgNames   []string
+}
+
+var (
+	MetaDataList = []MetaData{
```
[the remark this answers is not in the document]
iamniting: renamed it

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+
+	if err := r.reconcileOperators(); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.reconcileDynamicWatchers(req.Name); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.log.Info("ending reconcile")
```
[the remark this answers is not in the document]
iamniting: Done

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+
+	for _, metaData := range MetaDataList {
+
+		crList := &metav1.PartialObjectMetadataList{}
+		crList.TypeMeta.APIVersion = metaData.ApiVersion
+		crList.TypeMeta.Kind = metaData.Kind
+
+		err := r.Client.List(r.ctx, crList)
+		if err != nil {
+			if !meta.IsNoMatchError(err) {
```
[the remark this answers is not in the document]
iamniting: This is the case when kind is not registered. If we don't ignore this err we will always hit this again and again if the crd is not created for that kind.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+				msg := fmt.Sprintf("failed listing %s", metaData.Kind)
+				r.log.Error(err, msg)
+				multierr.AppendInto(&returnErr, err)
+			}
+		} else {
+			if len(crList.Items) >= 0 {
```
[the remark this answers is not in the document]
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	return nil
+}
+
+func (r *OperatorScalerReconciler) reconcileOperators() error {
+
+	var returnErr error
+
+	for _, metaData := range MetaDataList {
```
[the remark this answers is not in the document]
iamniting: Done

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+
+		crList := &metav1.PartialObjectMetadataList{}
+		crList.TypeMeta.APIVersion = metaData.ApiVersion
+		crList.TypeMeta.Kind = metaData.Kind
+
+		err := r.Client.List(r.ctx, crList)
```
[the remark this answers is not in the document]
iamniting: This func does not exists

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+				r.watching[crdName] = true
+			}
+		}
+	}
+
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *OperatorScalerReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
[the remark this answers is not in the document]
iamniting: I have kept the way they are in other files, Lets keep it as it is.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,333 @@
...
+				// Trigger a reconcile only during the creation of a specific CRD to ensure it runs exactly once for that CRD.
+				// This is required to dynamically add a watch for the corresponding Custom Resource (CR) based on the CRD name.
+				// The reconcile will be triggered with the CRD name as `req.Name`, and the reconciler will set up a watch for the CR using the CRD name.
+				createOnlyPredicate,
+				predicate.GenerationChangedPredicate{},
+			),
+		).
+		Build(r)
+
+	r.watching = make(map[string]bool)
```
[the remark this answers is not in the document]
iamniting: changed it

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
iamniting: I added the `client.Limit(1)`

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: We have other files where this function is the first. But regardless, please make sure all public functions are above all private functions.

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
nb-ohad: Yes, sorry for the confusion about the func name, it is `Limit`, not `Count`

controllers/operatorscaler_controller.go
nb-ohad: @iamniting You did not addressed the comment, you only renamed the function but the function still takes a list of CSVs

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+				predicate.NewPredicateFuncs(func(obj client.Object) bool {
+					return slices.ContainsFunc(ResourceMappingList, func(resourceMapping ResourceMappingRecord) bool {
+						return resourceMapping.CrdName == obj.GetName()
+					})
+				}),
+				// Trigger a reconcile only during the creation of a specific CRD to ensure it runs exactly once for that CRD.
+				// This is required to dynamically add a watch for the corresponding Custom Resource (CR) based on the CRD name.
+				// The reconcile will be triggered with the CRD name as `req.Name`, and the reconciler will set up a watch for the CR using the CRD name.
+				createOnlyPredicate,
+				predicate.GenerationChangedPredicate{},
```
nb-ohad, line 324: This is irrelevant, there cannot be a generation change if you are only watching create events

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+
+	return nil
+}
+
+func (r *OperatorScalerReconciler) reconcileOperators() error {
+
+	var returnErr error
+
+	for i := range ResourceMappingList {
+		resourceMapping := ResourceMappingList[i]
```
nb-ohad, line 189: You are taking a copy, not a reference
```suggestion
		resourceMapping := &ResourceMappingList[i]
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+
+		crList := &metav1.PartialObjectMetadataList{}
+		crList.TypeMeta.APIVersion = resourceMapping.ApiVersion
+		crList.TypeMeta.Kind = resourceMapping.Kind
+
+		err := r.Client.List(r.ctx, crList, client.Limit(1))
+		if err != nil {
```
nb-ohad, line 196: Prefer locality for error variables
```suggestion
		if err := r.Client.List(r.ctx, crList, client.Limit(1)); err != nil {
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+				err = r.Client.List(
+					r.ctx, csvList,
+					client.InNamespace(r.OperatorNamespace),
+					client.MatchingLabels(map[string]string{key: ""}),
+				)
+				if err != nil {
+					r.log.Error(err, "failed listing csvs with label", "label", key)
+					multierr.AppendInto(&returnErr, err)
+				} else {
+					err = r.updateCsvDeplymentsReplicas(csvList.Items...)
```
nb-ohad, line 218: As asked in the last review (https://github.com/red-hat-storage/odf-operator/pull/529#discussion_r1983419507), we should be looping over the CSVs here then sending each CSV into the func one by one.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+					if err != nil {
```
nb-ohad, line 219:
```suggestion
					if err = r.updateCsvDeplymentsReplicas(csvList.Items...); err != nil {
```

controllers/operatorscaler_controller.go
iamniting: @nb-ohad The way this func is written it can handle even a single entry, It will work for both of them.

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+func (r *OperatorScalerReconciler) updateCsvDeplymentsReplicas(csvList ...oprv1a1.ClusterServiceVersion) error {
+
+	var returnErr error
+
+	for idx := range csvList {
+		csv := &csvList[idx]
+		var updateRequired bool
+		for i := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
+			deploymentSpec := &csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[i].Spec
+			if *deploymentSpec.Replicas < int32(1) {
```
nb-ohad, line 240: There is no need to convert this to int32, comparison would work against a constant. 
```suggestion
			if *deploymentSpec.Replicas < 1 {
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+		err = fmt.Errorf("csv %s is not in succeeded state", OdfDepsSubscriptionStartingCSV)
+		r.log.Error(err, "waiting for csv to be in succeeded state")
+		return err
+	}
+
+	return nil
+}
+
+func (r *OperatorScalerReconciler) reconcileOperators() error {
+
```
nb-ohad, line 185: Can you please add an info log message here to record where we are in the reconcile process?
Maybe something like: "Entering reconcile operators"

Similar comment for the other 2 main functions of the reconcile

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+					if err != nil {
+						r.log.Error(err, "failed updating csvs replica")
+						multierr.AppendInto(&returnErr, err)
+					}
+				}
+			}
+		}
+	}
+
+	return returnErr
```
nb-ohad, line 228: Can you please add an info log message here to record where we are in the reconcile process? 
You should have only in cases where we don't return an error:
Maybe something like: "Reconcile operators completed successfully"

Similar comment for the other 2 main functions of the reconcile

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+
+		if !r.kindsBeingWatched[resourceMapping.Kind] {
+
+			crd := &extv1.CustomResourceDefinition{}
+			if err := r.Client.Get(r.ctx, types.NamespacedName{Name: resourceMapping.CrdName}, crd); err != nil {
+				if !errors.IsNotFound(err) {
+					r.log.Error(err, "failed getting crd", "crdName", resourceMapping.CrdName)
+					return err
+				}
+			} else {
```
nb-ohad, line 272: There is no need to for the added nesting. You can use this pattern instead
```suggestion
			crd := &extv1.CustomResourceDefinition{}
			crs.Name =  resourceMapping.CrdName
			if err := r.Client.Get(r.ctx, client.ObjectKeyFromObject(crs), crd); client.IgnoreNotFound(err) != nil {
				r.log.Error(err, "failed getting crd", "crdName", resourceMapping.CrdName)
				return err
			} else if err == nil {
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,334 @@
...
+									APIVersion: resourceMapping.ApiVersion,
+									Kind:       resourceMapping.Kind,
+								},
+							},
+						),
+						&handler.EnqueueRequestForObject{},
+						// Trigger the reconcile for creation events of the object.
+						// This ensures the replicas in the CSV are scaled up based on the presence of Custom Resource (CR).
+						createOnlyPredicate,
+						predicate.GenerationChangedPredicate{},
```
nb-ohad, line 290: This predicate is irrelevant if we only consider create events. 
```suggestion
```

nb-ohad (changes requested):

controllers/operatorscaler_controller.go
iamniting: I have replied there.

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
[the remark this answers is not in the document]
iamniting: created a new map as discussed.

controllers/operatorscaler_controller.go
iamniting: Added a loop here

controllers/operatorscaler_controller.go
iamniting: Changed it to `successfully completed *`

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,329 @@
...
+
+	r.log.Info("successfully completed reconcileDynamicWatchers")
+	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *OperatorScalerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	// It will keep track or the crds that we care about
+	crdMap := make(map[string]bool)
```
nb-ohad, line 302: There is no need for a make here, the type is hard coded
```suggestion
	crdMap := map[string]bool{}
```

controllers/operatorscaler_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	for _, resMap := range ResourceMappingList {
+		crdMap[resMap.CrdName] = true
+	}
```
nb-ohad, line 305: There is no need to create a copy of each item while creating the map
```suggestion
	for i := range ResourceMappingList {
		crdName := ResourceMappingList[i]
		crdMap[crdName] = true
	}
```

nb-ohad (changes requested):

controllers/operatorscaler_controller.go
iamniting: Done

controllers/operatorscaler_controller.go
iamniting: Done

iamniting:
Things are working as expected, I am removing the hold
/unhold

PR #552, opened 2025-03-13
remove storagesystem controller

iamniting:


api/v1alpha1/storagesystem_types.go
nb-ohad, line 1: Why do we need to keep anything in this file or the file at all?
We just need to made sure the OpenAPI representation of this API is in the operator's bundle and this can be achieved even without have a go structs for the API

nb-ohad (changes requested):

api/v1alpha1/storagesystem_types.go
iamniting: We still need to watch, read, and delete via the other cleanup controller. Thats why we need it.

nb-ohad: But you don't need the types. You just need the OpenAPI yaml file (CRD file). Can't we get rid of this types file?

nb-ohad:
@Nikhil-Ladha I gave this an LGTM, but I still would like to have an answer for: https://github.com/red-hat-storage/odf-operator/pull/552#discussion_r2010951963 and if the answer is that we don't need this file I would expect a follow up PR

iamniting:
> @Nikhil-Ladha I gave this an LGTM, but I still would like to have an answer for: [#552 (comment)](https://github.com/red-hat-storage/odf-operator/pull/552#discussion_r2010951963) and if the answer is that we don't need this file I would expect a follow up PR

I guess you mean @iamniting

api/v1alpha1/storagesystem_types.go
iamniting: How will we read the storage system and its specs in the controller if we remove the go struct? How the cleanup controller will watch the storagesystem without the go struct?

PR #553, opened 2025-03-13
ci: always print the logs in case of failure or success in the CI

iamniting:

### DIFF
--- a/hack/install-odf.sh
+++ b/hack/install-odf.sh
@@ -9,8 +9,22 @@ OPERATOR_SDK=${OPERATOR_SDK:-$1}
 BUNDLE_IMG=${BUNDLE_IMG:-$2}
 ODF_DEPS_CATALOG_IMG=${ODF_DEPS_CATALOG_IMG:-$3}
 CSV_NAMES=${CSV_NAMES:-$4}
+CI=${CI:-false}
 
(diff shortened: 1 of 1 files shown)


hack/install-odf.sh
```diff
@@ -9,8 +9,22 @@ OPERATOR_SDK=${OPERATOR_SDK:-$1}
...
+
+function print_debug_logs {
+    if [ "$CI" == true ]; then
+        echo "printing debug logs"
+        oc get csv -n "$INSTALL_NAMESPACE"
+        oc get pods -n "$INSTALL_NAMESPACE"
+        oc describe csv -n "$INSTALL_NAMESPACE"
+        oc describe pods -n "$INSTALL_NAMESPACE"
+    fi
+}
```
umangachapagain, line 24: Why not use the CI feature to run must-gather to collect logs after tests?

iamniting: We are already collecting must gather. But having it here is useful. Because in case of failure, we see this directly on the job page instead of going here and there and downloading the must gather.

umangachapagain: It's simple enough, but if you wish to print to stdout, it's your call.

malayparida2000 (approved):
/lgtm

PR #555, opened 2025-03-14
Move the creation of metrics to operatorScaler controllers

iamniting:

Related PR: https://github.com/red-hat-storage/ocs-operator/pull/3090

### DIFF
--- a/controllers/operatorscaler_controller.go
+++ b/controllers/operatorscaler_controller.go
@@ -19,9 +19,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/red-hat-storage/odf-operator/metrics"
 	"go.uber.org/multierr"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -149,6 +151,10 @@ func (r *OperatorScalerReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	if err := r.reconcileMetrics(ctx, logger); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	logger.Info("reconcile completed successfully")
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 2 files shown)


controllers/operatorscaler_controller.go
```diff
@@ -144,6 +149,10 @@ func (r *OperatorScalerReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	if err := r.createMetrics(); err != nil {
```
leelavg, line 152: possible to stick to naming this as `reconcile...`?

controllers/operatorscaler_controller.go
```diff
@@ -177,6 +186,48 @@ func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady() error {
 	return nil
 }
 
+func (r *OperatorScalerReconciler) createMetrics() error {
+	r.log.Info("entering createMetrics")
```
leelavg, line 190: i don't think it's good to couple this log w/ the function name and also no need to mark the location, I see it's being done in other functions but not a good idea.

i don't see the need for logging here at all, if required i'd put a log for `reporting metrics for xyz` in the else.

controllers/operatorscaler_controller.go
```diff
@@ -177,6 +186,48 @@ func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady() error {
...
+				if !meta.IsNoMatchError(err) {
+					msg := fmt.Sprintf("failed listing %s", resMap.Kind)
+					r.log.Error(err, msg)
+					multierr.AppendInto(&returnErr, err)
+				}
+			} else if err == nil {
+				for j := range crList.Items {
+					crItem := &crList.Items[j]
+
+					metrics.ReportODFSystemMapMetrics(
```
leelavg, line 213: may I know what type of prom metrics is this? are we fine w/ just exporting this during reconcile?

controllers/operatorscaler_controller.go
```diff
@@ -47,19 +49,21 @@ const (
...
 type ResourceMappingRecord struct {
-	CrdName    string
-	ApiVersion string
-	Kind       string
-	PkgNames   []string
+	CrdName       string
+	ApiVersion    string
+	Kind          string
+	PkgNames      []string
+	ReportMetrics bool
```
leelavg, line 56: I don't think it's good to piggyback on this, I believe the original intention was this record will be supplied to the operator, better to have this if it makes sense for all the records.

controllers/operatorscaler_controller.go
iamniting: This is the same metrics that UI folks were talking about, and even earlier, we were reporting inside the reconcile.

controllers/operatorscaler_controller.go
iamniting: Maybe in future, we want to report this metrics for other storage systems also. Thats why this should be a good place IMO.

controllers/operatorscaler_controller.go
iamniting: I can do that.

iamniting: Done

controllers/operatorscaler_controller.go
iamniting: It was requested here https://github.com/red-hat-storage/odf-operator/pull/529#discussion_r1986921384

controllers/operatorscaler_controller.go
iamniting: I removed it and switched to the fixed list.

controllers/operatorscaler_controller.go
```diff
@@ -174,6 +180,52 @@ func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady(ctx context.Context
 	return nil
 }
 
+func (r *OperatorScalerReconciler) reconcileMetrics(ctx context.Context, logger logr.Logger) error {
+	logger.Info("entering reconcileMetrics")
+
+	var returnErr error
```
nb-ohad, line 186: 1. Can this please be moved to the area where it is used? direct above the for loop? 
2. This is not the only error var the might get returned so returnErr does not fit as a name, maybe aggrError is more fitting

controllers/operatorscaler_controller.go
```diff
@@ -174,6 +180,52 @@ func (r *OperatorScalerReconciler) isOdfDependenciesCsvReady(ctx context.Context
...
+				crItem := &crList.Items[j]
+
+				metrics.ReportODFSystemMapMetrics(
+					crItem.Name+"-storagesystem",
+					crItem.Name,
+					crItem.Namespace,
+					strings.ToLower(crList.Kind)+crList.APIVersion,
+				)
+			}
+		}
```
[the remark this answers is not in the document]
iamniting: Actually we dont need meta.isnomatch error as we are listing the crds and not crs.

controllers/operatorscaler_controller.go
iamniting: Moved it.

PR #557, opened 2025-03-14
controllers: remove the storagecluster controller

iamniting:
It was created to watch the storagecluster and create the storagesystem for that storagecluster.

As storagesystem is being removed we don't need this controller.


controllers/storagesystem_controller_fake.go
nb-ohad, line 1: What is this file and what is it used for?

nb-ohad (changes requested):

controllers/storagesystem_controller_fake.go
iamniting: This file was used for getting some constants in the tests. We should have renamed this with _test.go. Anyway we are removing it now.

PR #565, opened 2025-03-20
switch to dynamic mapping instead of using hardcoded mapping

iamniting:
Start using the configuration from the configmap which allows the user to change the configuration at any time.

### DIFF
--- a/controllers/config.go
+++ b/controllers/config.go
@@ -1,0 +1,235 @@
+/*
+Copyright 2025 Red Hat OpenShift Data Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/ghodss/yaml"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	odfOperatorConfigMapName = "odf-operator-manager-config"
+)
+
+var (
+	// configMapResorceVersion will keep the local copy of the odf-operator-manager-config configmap resource version
+	configMapResorceVersion string
+	// OlmPkgRecordList will keep the local copy of parsed data from odf-operator-manager-config configmap
+	OlmPkgRecordList = []*OlmPkgRecord{}
+	// CrdToPackageRecordList will keep the local copy of parsed data from OlmPkgRecordList
+	CrdToPackageRecordList = []*CrdToPackageRecord{}
+)
+
+type OlmPkgRecord struct {
+	/* example
+	   channel: alpha
+	   csv: ocs-operator.v4.18.0
+	   pkg: ocs-operator
+	   scaleUpOnInstanceOf:
+	     - storageclusters.ocs.openshift.io
+	   ---------------------------------------
+	   channel: beta
+	   csv: odf-prometheus-operator.v4.18.0
+	   pkg: odf-prometheus-operator
+	   scaleUpOnInstanceOf:
+	     - alertmanagers.monitoring.coreos.com
+	     - prometheuses.monitoring.coreos.com
+	*/
+
+	Channel             string   `yaml:"channel"`
+	Csv                 string   `yaml:"csv"`
+	Pkg                 string   `yaml:"pkg"`
+	ScaleUpOnInstanceOf []string `yaml:"ScaleUpOnInstanceOf"`
+}
+
+type CrdToPackageRecord struct {
+	/* examples
+	   CrdName:    "storageclusters.ocs.openshift.io",
+	   ApiVersion: "ocs.openshift.io/v1",
+	   Kind:       "StorageCluster",
+	   PkgNames:   []string{OcsSubscriptionPackage},
+	   -----------------------------------------------
+	   CrdName:    "cephclusters.ceph.rook.io",
+	   ApiVersion: "ceph.rook.io/v1",
+	   Kind:       "CephCluster",
+	   PkgNames:   []string{RookSubscriptionPackage, CephCSISubscriptionPackage, CSIAddonsSubscriptionPackage},
+	*/
+
+	CrdName    string
+	ApiVersion string
+	Kind       string
+	PkgNames   []string
+}
+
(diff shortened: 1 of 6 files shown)


controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+	PkgNames   []string
+}
+
+func GetCrdAndPackagesMappingList(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) ([]CrdAndPackages, error) {
+
+	configMapData, err := GetConfig(ctx, cli, logger, operatorNamespace)
+	if err != nil {
+		logger.Error(err, "failed to get configmap data")
+		return nil, err
+	}
```
nb-ohad, line 117: This need to be provided from the outside as there is no point in reading and parsing the same source map in multiple places

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+				logger.Error(err, "CRD not found", "crdName", record.CrdName)
+				continue
+			}
+			logger.Error(err, "failed to get CRD", "crdName", record.CrdName)
+			return nil, err
+		}
+		resourceMapping[i].CrdPresent = true
+		resourceMapping[i].ApiVersion = crd.Spec.Group + "/" + crd.Spec.Versions[0].Name
+		resourceMapping[i].Kind = crd.Spec.Names.Kind
+	}
```
nb-ohad, line 151: This logic is over-complicated without any real reason to be
You start with a list of package records that embed a list of crd of interest and want to get to a list of crd records with some extra information (kind and API version) and a sub list of tied packages.

There is no need to loop 3 separate times, you loop once and build each record whenever you see a new crd name or update an existing record whenever you see a crd name that already exists.

Adding a high-level pseudo code to describe a simpler solution
```
crdMapping = map[string]*crdMappingRecord{}

for pkgInfo := range packageList {
    for crdName := range pkgInfo.scalerCrds {
        if crdInfo :=  crdMapping[crdName]; crdInfo == nil {
               client.Get(CustomResourceDefinition)
               crdInfo = crdMappingRecord{
                   CrdName = crdName,
                   Kind = ... 
                   ApiVersion = ...
                   PkgNames = []string{}
               }
        }
        crdInfo.PkgNames = append(crdInfo.PkgNames, pkgInfo.Name)
    }
}

```

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+	return resourceMapping, nil
+}
+
+func GetCrdNamesMap(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) (map[string]bool, error) {
+
+	configMapData, err := GetConfig(ctx, cli, logger, operatorNamespace)
+	if err != nil {
+		logger.Error(err, "failed to get configmap data")
+		return nil, err
+	}
```
nb-ohad, line 164: This need to be provided from the outside as there is no point in reading and parsing the same source map in multiple places

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+		resourceMapping[i].ApiVersion = crd.Spec.Group + "/" + crd.Spec.Versions[0].Name
+		resourceMapping[i].Kind = crd.Spec.Names.Kind
+	}
+
+	logger.Info("parsed resource mapping successfully", "resourceMapping", resourceMapping)
+
+	return resourceMapping, nil
+}
+
+func GetCrdNamesMap(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) (map[string]bool, error) {
```
nb-ohad, line 158: Where is this used and why do we need a map and not a list ?

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+		logger.Error(err, "failed to get configmap data")
+		return nil, err
+	}
+
+	crdNames := map[string]bool{}
+	for _, value := range configMapData {
+		for _, crd := range value.ScalerCrds {
+			crdNames[crd] = true
+		}
+	}
```
nb-ohad, line 171: Don't create a copy of struct values
```suggestion
	for i := range configMapData {
		for _, crdName := range configMapData[i].ScalerCrds {
			crdNames[crdName] = true
		}
	}
```

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+
+	logger.Info("parsed CRD names successfully", "crdNames", crdNames)
+
+	return crdNames, nil
+}
+
+func GetCsvNamesMap(ctx context.Context, cli client.Client, logger logr.Logger, operatorNamespace string) (map[string]bool, error) {
```
nb-ohad, line 178: Where is this used and why do we need a map and not a list ?

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+
+	configMapData, err := GetConfig(ctx, cli, logger, operatorNamespace)
+	if err != nil {
+		logger.Error(err, "failed to get configmap data")
+		return nil, err
+	}
```
nb-ohad, line 184: This need to be provided from the outside as there is no point in reading and parsing the same source map in multiple places

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+
+	csvNames := map[string]bool{}
+	for _, value := range configMapData {
+		csvNames[value.Csv] = true
+	}
```
nb-ohad, line 189: I don't understand what do we get by creating this map. There is no flatting operation here and we are talking about a small list (up to 20 items) where is make more sense to just use Contains

nb-ohad (changes requested):

controllers/config.go
```diff
@@ -0,0 +1,231 @@
...
+	cm := &corev1.ConfigMap{}
+	cm.Name = odfOperatorConfigmapName
+	cm.Namespace = operatorNamespace
+	if err := cli.Get(ctx, client.ObjectKeyFromObject(cm), cm); err != nil {
+		logger.Error(err, "failed to get configmap")
+		return nil, err
+	}
+
+	// parse the ConfigMap data and skip any keys that fail to parse
+	configMapData := make(map[string]ConfigData)
```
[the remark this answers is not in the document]
iamniting: It gives us the flexibility even to read the key if we are interested in any key maybe in the future.

controllers/config.go
iamniting: It is a pointer to a struct. So we are not creating a copy.

controllers/config.go
iamniting: This is used in the webhook to compare whether the CSV is ODF-owned.

controllers/config.go
iamniting: What is a problem with map, when we can get a O(1) search operation.

controllers/config.go
```diff
@@ -0,0 +1,235 @@
...
+	odfOperatorConfigMapName = "odf-operator-manager-config"
+)
+
+var (
+	// configMapResorceVersion will keep the local copy of the odf-operator-manager-config configmap resource version
+	configMapResorceVersion string
+	// OlmPkgRecordList will keep the local copy of parsed data from odf-operator-manager-config configmap
+	OlmPkgRecordList = []*OlmPkgRecord{}
+	// CrdToPackageRecordList will keep the local copy of parsed data from OlmPkgRecordList
+	CrdToPackageRecordList = []*CrdToPackageRecord{}
```
leelavg, line 41: after going through the file, you are creating empty slice here and assign a new slice after copying the details to an intermediate slice which leaves old slice for gc everytime.

I propose 2 things, one, the structs doesn't nest with pointers and so you can store the structs directly (only one slice to be pedantic), two, update the older slice with intermediate info and cut down the extra space.

iamniting:
Closing this, This is done by #588

PR #574, opened 2025-03-25
remove the unused apis

iamniting:


hack/make-project-vars.mk
```diff
@@ -5,7 +5,7 @@ ENVTEST_ASSETS_DIR := $(PROJECT_DIR)/testbin
 GOBIN ?= $(BIN_DIR)
 GOOS ?= $(shell go env GOOS)
 GOARCH ?= $(shell go env GOARCH)
-GOPROXY ?= direct,https://proxy.golang.org
+GOPROXY ?= https://proxy.golang.org/
```
leelavg, line 8: atleast add `direct` to the end, this might fail if the pkg isn't found the proxy afair

controllers/quickstarts_test.go
```diff
@@ -20,13 +20,13 @@ import (
...
 	"testing"
 
-	"github.com/ghodss/yaml"
 	consolev1 "github.com/openshift/api/console/v1"
 	"github.com/stretchr/testify/assert"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/yaml"
```
leelavg, line 29: so this provides the same semantics as removed pkg w/ no changes?

leelavg (commented):
cool, so much is removed.

controllers/quickstarts_test.go
iamniting: yes, i verified that, having one package in the tests and another in the actual code.

hack/make-project-vars.mk
iamniting: No, it never failed; it was the same earlier. First of all, if you add something to this module, you will make the changes on your local machine, and pulling the new module on your local machine will make it available on the proxy server, provided you are the first one to pull it. Otherwise it will be already present on the proxy server.

PR #578, opened 2025-03-27
controllers: consolidate all subs and CSVs under single management

iamniting:
previously, subscriptions were created per StorageSystem, resulting in fragmented subscription handling. With the deprecation of StorageSystem APIs, we now treat all subscriptions uniformly.

This change:
- combines all subscriptions and CSVs under centralized logic
- removes StorageSystem-specific subscription handling
- ensures consistent subscription behavior across the cluster
- auto-approve install plans for subscribed CSVs only


controllers/subscriptions.go
```diff
@@ -230,63 +228,83 @@ func GetOdfSubscription(cli client.Client) (*operatorv1alpha1.Subscription, erro
...
+	return nil
+}
+
+func isSubscriptionPresentForCsv(cli client.Client, csvName string) (bool, error) {
+
+	// find the package name for the given csv
+	var pkgName string
+	subs := GetSubscriptions()
+	for _, sub := range subs {
+		if sub.Spec.StartingCSV == csvName {
```
nb-ohad, line 288: I am not sure you will always have a starting CSV. Also the starting CSV does not mean that actual CSV. 
An example: Assume the subscription starting CSV is pointing to an older CSV, but the install plan for that older CSV already got approved so the operator already moved to the next release (not the target release). In that case you will not find the starting CSV anymore

nb-ohad (changes requested):

iamniting:
CI failed again due to provisioning. I tested the change locally, and it worked. Overriding the CI.

PR #581, opened 2025-04-01
Switch to dynamic mapping for the subscription controller

iamniting:

Depends on #580

### DIFF
--- a/controllers/subscription_controller.go
+++ b/controllers/subscription_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/operator-framework/operator-lib/conditions"
 	"go.uber.org/multierr"
 	admrv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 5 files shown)


controllers/subscription_controller.go
```diff
@@ -67,15 +68,29 @@ type SubscriptionReconciler struct {
...
+	configmap, err := GetOdfConfigMap(ctx, r.Client, logger)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 
-	if err := r.setOperatorCondition(logger, olmPkgRecords); err != nil {
+	var olmPkgRecords []*OlmPkgRecord
+	var csvNames []string
+	ParseOdfConfigMapRecords(logger, configmap, func(record *OlmPkgRecord) {
+		if record.Pkg != "" && record.Csv != "" && record.Channel != "" {
```
nb-ohad, line 86: Why do we need this check? And if we do I would suggest to at least report something in the log saying that we are skipping a specific key

controllers/subscription_controller.go
```diff
@@ -67,15 +68,29 @@ type SubscriptionReconciler struct {
...
+			csvNames = append(csvNames, record.Csv)
+			olmPkgRecords = append(olmPkgRecords, record)
+		}
+	})
+	logger.Info("subscriptions records", "olmPkgRecords", olmPkgRecords)
```
nb-ohad, line 91: I would say this is more of a debug info then an "info" into. Can we log this using debug level?

controllers/subscription_controller.go
```diff
@@ -118,17 +134,17 @@ func (r *SubscriptionReconciler) ensureSubscriptions(logger logr.Logger, olmPkgR
...
 		logger.Error(err, "failed to list OperatorConditions")
 		return err
 	}
 
-	condMap := make(map[string]struct{}, len(olmPkgRecords))
-	for _, olmPkgRecord := range olmPkgRecords {
-		condMap[olmPkgRecord.Csv] = struct{}{}
+	condMap := make(map[string]struct{}, len(csvNames))
+	for _, csvName := range csvNames {
+		condMap[csvName] = struct{}{}
```
nb-ohad, line 147: If we need a map/set for this controller, there is no point in reading the config map CSV names into a slice then converting into a map/set. You should be reading directly into a map/set

controllers/subscription_controller.go
```diff
@@ -242,6 +258,16 @@ func (r *SubscriptionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&handler.EnqueueRequestForObject{},
 			builder.WithPredicates(predicate.GenerationChangedPredicate{}),
 		).
+		Watches(
+			&corev1.ConfigMap{},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(
+				predicate.NewPredicateFuncs(func(obj client.Object) bool {
+					return obj.GetName() == odfOperatorConfigMapName && obj.GetNamespace() == r.OperatorNamespace
+				}),
```
nb-ohad, line 267: I think we had a util function to create a predicate for a give NamespacedName. Maybe I am confusing it with a different repo

nb-ohad (changes requested):

controllers/subscription_controller.go
iamniting: We dont have it here.

controllers/subscription_controller.go
iamniting: If there is a mistake in the keys of the parsed data, we should skip that, as using it will have unexpected results. Added the loggging.

controllers/subscription_controller.go

controllers/subscription_controller.go
iamniting: We don't have Debug. We can evaluate this as a tech debt.

PR #584, opened 2025-04-02
update deps bundles version

iamniting:


bundle/odf-dependencies/metadata/dependencies.yaml
```diff
@@ -2,31 +2,31 @@ dependencies:
 - type: olm.package
   value:
     packageName: ocs-operator
-    version: ">=4.17.0 <=4.18.0"
+    version: "4.19.0"
```
Nikhil-Ladha, line 5: Do we have a breaking change, for limiting the value to 4.19 and not giving a range?

iamniting: We wanted to lock the versions to the exact version for more control. This also allows us to upgrade everything at once. We will not see issues where one operator is upgraded and other one is not.

Nikhil-Ladha (approved):

PR #586, opened 2025-04-02
controllers: use configmap to get the csv names for webhook

iamniting:

Depends on #581

### DIFF
--- a/webhook/csv.go
+++ b/webhook/csv.go
@@ -21,7 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strings"
+	"sync"
 
 	"github.com/go-logr/logr"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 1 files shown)


webhook/csv.go
```diff
@@ -40,9 +40,14 @@ type ClusterServiceVersionDeploymentScaler struct {
 
 	Decoder           admission.Decoder
 	OperatorNamespace string
+
+	mutex                               sync.Mutex
+	odfOperatorConfigMapResourceVersion string
+	odfOwnedCsvNames                    map[string]bool
```
nb-ohad, line 46:
```suggestion
	odfOwnedCsvSet                   map[string]bool
```

webhook/csv.go
```diff
@@ -40,9 +40,14 @@ type ClusterServiceVersionDeploymentScaler struct {
 
 	Decoder           admission.Decoder
 	OperatorNamespace string
+
+	mutex                               sync.Mutex
```
nb-ohad, line 44: `mutex` is such a generic name. proposing: odfOpeeratorConfigAccessMutext

webhook/csv.go
```diff
@@ -82,6 +92,34 @@ func (r *ClusterServiceVersionDeploymentScaler) Handle(ctx context.Context, req
...
+	}
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if configmap.ResourceVersion == r.odfOperatorConfigMapResourceVersion {
+		return nil
+	}
+
+	r.odfOwnedCsvNames = map[string]bool{}
```
nb-ohad, line 109: I was under the impression that you like to implement sets  using `map[string]struct{}`

nb-ohad (changes requested):

webhook/csv.go
iamniting: Here, it makes sense to use the bool so we can return directly without checking the existence of the key.

webhook/csv.go
iamniting: Done

webhook/csv.go
iamniting: It depends on the usecase.

iamniting:
/override ci/prow/ci-bundle-odf-dependencies-catalog
/override ci/prow/images
/override ci/prow/ci-bundle-odf-operator-bundle

The tests passed earlier.

PR #592, opened 2025-04-03
controllers: remove the usage of old config variables

iamniting:

Depends on #588


malayparida2000:
/lgtm 
As Ohad is not around & as mentioned by Nitin that it's just a rebase & push.

PR #598, opened 2025-04-04
controllers: fix replica update logic for CSVs during upgrades

iamniting:
Previously, the controller would list all CSVs with a package label and update their replicas. This caused issues during upgrades because:

- It would scale up the ocs-client-operator 4.18 CSV.
- The ocs-client-operator 4.18 CSV pods would get stuck in init state.
- This blocked the upgrade of ocs-client-operator the 4.18 CSV remained in Installing state while 4.19 CSV was stuck in Pending.

The new implementation specifically targets the correct CSV to avoid interfering with Previous CSV during upgrades.

This occurred because odf-operator 4.19 gets upgraded before other operators and start modifying the 4.18 CSVs.

The problem is happening only with the ocs-client-operator as it was scaled down by odf-operator in 4.18 and previos releases.

Bug: https://issues.redhat.com/browse/DFBUGS-2069

### DIFF
--- a/controllers/operatorscaler_controller.go
+++ b/controllers/operatorscaler_controller.go
@@ -45,11 +45,6 @@ import (
 	"github.com/red-hat-storage/odf-operator/metrics"
 )
 
-const (
-	// csvLabelKey is a label key to identify CSV with pkg name and namespace
-	CsvLabelKey = "operators.coreos.com/%s.%s"
-)
-
 var (
 	createOnlyPredicate = predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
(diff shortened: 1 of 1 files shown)


leelavg:
What is better? A custom code for 4.19 which we can delete in 4.20 or targetting csv everytime (as implemented in this PR)?

The only diff would be, we will ask during build time to mention the CSV rather than Pkg name or am I missing anything else? Being exact is good, I'm just thinking if we are restricting any agreed upon features by changing this.

iamniting:
> What is better? A custom code for 4.19 which we can delete in 4.20 or targetting csv everytime (as implemented in this PR)?
> 
> The only diff would be, we will ask during build time to mention the CSV rather than Pkg name or am I missing anything else? Being exact is good, I'm just thinking if we are restricting any agreed upon features by changing this.

We already have pkg, csv, and channel defined in the ConfigMap, so we don't need to repopulate it. Converting the pkg to CSV will cause the subscription controller's functionality issues.

leelavg:
> the pkg to CSV will cause the subscription controller's functionality issues.
- ah! this is due to us using label name w/ pkg for listing CSVs, now there is no use for pkg I guess?
- BTW, PR is addressing a generic scenario, isn't? Not to touch older CSVs that were intentionally scaled down and that creates a divide b/n what odf-op was acting on in previous config vs now?

What I'm expecting is a rejection list for CSV not to scale up as from 4.19 if someone wants to scale down a certain operator/csv they can't do that unless they scale down odf-op. Thinking whether this happens in production or not, certainly useful during development/testing. 

Anyways, will discuss on Monday or Sunday when Ohad comes online depending on your availability, don't want to eat away your weekend anymore 😀.

nb-ohad:
This changes seems reasonable but the parsed struct name need to change from `KindPackagesRecord` to `KindCsvsRecord`. I am LGTMing this PR with the expectation to get the rename on a future PR

PR #600, opened 2025-04-05
controllers: fix upgrade deadlock in cleanup controller

iamniting:
During an ODF operator upgrade, a race condition occurs where:
- odf-operator upgrades first to 4.19 and removes the StorageSystem ownerReference from StorageCluster
- ocs-operator (still at v4.18) fails to find this ownerReference, marking itself as not upgradable

This creates a deadlock because:
- ocs-operator requires the StorageSystem ownerReference to mark itself upgradable
- odf-operator has already deleted both the ownerReference and StorageSystem CR

The fix ensures that ocs-operator CSV upgrades before running the cleanup, preventing this race condition.

### DIFF
--- a/controllers/cleanup_controller.go
+++ b/controllers/cleanup_controller.go
@@ -22,6 +22,7 @@ import (
 	"slices"
 
 	"github.com/go-logr/logr"
+	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


nb-ohad:
This PR implements a generalized fix that we don't want to carry forward. As discussed offline, we will created a more specific and leaner fix for the same problem. At that point this PR will get closed in favor of the new PR. 

**Please do not LGTM or remove the hold from this PR**

iamniting:
It is done with the #601

PR #601, opened 2025-04-07
controllers: fix upgrade deadlock in cleanup controller 

leelavg:
During an ODF operator upgrade, a race condition occurs where:
- odf-operator upgrades first to 4.19 and removes the StorageSystem
  ownerReference from StorageCluster
- ocs-operator (still at v4.18) fails to find this ownerReference,
  marking itself as not upgradable

This creates a deadlock because:
- ocs-operator requires the StorageSystem ownerReference to mark
  itself upgradable
- odf-operator has already deleted both the ownerReference and
  StorageSystem CR

The fix ensures that ocs-operator CSV upgrades before running the
cleanup, preventing this race condition.

Issue is identified and initial fix was provided by @iamniting.


leelavg:
supersedes #600

controllers/cleanup_controller.go
```diff
@@ -66,6 +69,11 @@ func (r *CleanupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	logger.Info("storagesystem instance found")
 
+	// TODO: this check should be removed in 4.20
```
nb-ohad, line 72: Please add the context and need for this check on the comment

controllers/cleanup_controller.go
```diff
@@ -74,6 +82,28 @@ func (r *CleanupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		return fmt.Errorf("failed to list subscriptions: %v", err)
+	}
+	ocsOperatorSubscription := util.Find(subscriptionList.Items, func(sub *opv1a1.Subscription) bool {
+		return sub.Spec.Package == "ocs-operator"
+	})
+	if ocsOperatorSubscription == nil {
+		return fmt.Errorf("no subscription exists with package 'ocs-operator'")
+	} else if strings.HasSuffix(ocsOperatorSubscription.Spec.Channel, "4.18") {
+		return fmt.Errorf("subscription of 'ocs-operator' still points to '4.18'")
+	} else if !strings.HasSuffix(ocsOperatorSubscription.Spec.Channel, "4.19") {
```
nb-ohad, line 97: Why are we using a suffix check here? what will happen with dev/test builds?

nb-ohad (changes requested):

controllers/cleanup_controller.go
leelavg: The channel is usually prefixed w/ `stable-`, this doesn't effect dev testing as upgrades is validated w/ d/s images.

controllers/cleanup_controller.go
leelavg: done.

PR #609, opened 2025-04-14
controllers: Wait for OCS CSV to be fully ready before cleanup

iamniting:
The Subscription's currentCSV field gets updated immediately when a new CSV is created, even when the new CSV is still in 'Pending' state and the old CSV is in 'Replacing' state. This leads to premature cleanup because the old ocs-operator is still running.

Modify the cleanup logic to only proceed when:
1. The new CSV reaches 'Succeeded' phase

### DIFF
--- a/controllers/cleanup_controller.go
+++ b/controllers/cleanup_controller.go
@@ -78,7 +78,7 @@ func (r *CleanupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	//  odf-operator has already deleted both the ownerReference and StorageSystem CR
 	// The function ensures that ocs-operator CSV upgrades before running the cleanup, preventing this race condition.
 	// In 4.19 we are not using StorageSystem CR, so the check should be removed in 4.20
-	if err := r.isOcsOperatorSubAndCSVAt419(ctx); err != nil {
+	if err := r.isOcsOperatorSubAndCSVAt419(ctx, logger); err != nil {
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


controllers/cleanup_controller.go
```diff
@@ -103,12 +103,34 @@ func (r *CleanupReconciler) isOcsOperatorSubAndCSVAt419(ctx context.Context) err
...
+
+	err := r.Client.Get(ctx, client.ObjectKeyFromObject(ocsCsv), ocsCsv)
+	if err != nil {
+		logger.Error(err, "failed getting ocs csv", "csvName", ocsCsv)
+		return err
+	}
+
+	if ocsCsv.Status.Phase != opv1a1.CSVPhaseSucceeded {
+		err = fmt.Errorf("csv %s is not in succeeded state", ocsCsvName)
+		logger.Error(err, "waiting for csv to be in succeeded state")
```
leelavg, line 130: no new error here, the error returned here anyways shows up in reconciler. so, no need to explicitly log in here.

iamniting: Reconciler errors are logged after the full reconcile call is completed. It is possible that a few more logs will be populated depending on the error handling, so it is better to log the error where it occurs for ease of debugging.

leelavg (approved):

iamniting:
/override ci/prow/ci-bundle-odf-operator-bundle
/override ci/prow/images
/override ci/prow/ci-bundle-odf-dependencies-catalog

This controller is not being tested in the tests, So their is no point to test the e2e.

PR #613, opened 2025-05-08
DFBUGS-2486:[release-4.18] update ibm odf operator to 1.7.1

kasserater:
use IBM ODF v1.7.1, which uses the same binaries of v1.7.0, but changes the dependency on CSI 1.12.2 from EXACT version to AT LEAST version


hack/make-bundle-vars.mk
```diff
@@ -106,7 +106,7 @@ CEPHCSI_BUNDLE_IMG ?= $(CEPHCSI_BUNDLE_IMG_LOCATION)/$(CEPHCSI_BUNDLE_IMG_NAME):
 
 IBM_BUNDLE_NAME ?= ibm-storage-odf-operator
 IBM_BUNDLE_IMG_NAME ?= $(IBM_BUNDLE_NAME)-bundle
-IBM_BUNDLE_VERSION ?= 1.7.0
+IBM_BUNDLE_VERSION ?= 1.7.1
 IBM_BUNDLE_IMG_TAG ?= 1.7.0
```
iamniting, line 110: Why are we using the same tag 1.7.0 for 1.7.1? We should use 1.7.1 IMO.

iamniting (changes requested):
Can you pls change the commit msg? It should not exceed more than 73 chars.

something like `bundle: update ibm odf version from 1.7.0 to 1.7.1`

hack/make-bundle-vars.mk
iamniting: It looks like (after looking at the failure) you forgot to change this. Can you pls change it?

iamniting:
@kasserater We will also require a bug to merge this PR. Can you pls create a bug? Bugs will be required for the other branches as well.

hack/make-bundle-vars.mk
kasserater: it must be the same - we are using the same binaries, only changing the dependencies.yaml of the release

kasserater:
> @kasserater We will also require a bug to merge this PR. Can you pls create a bug? Bugs will be required for the other branches as well.

https://issues.redhat.com/browse/DFBUGS-2486

hack/make-bundle-vars.mk
iamniting: The image still points to the old 1.7.0 bundle.

kasserater:
@openshift-ci-robot /retest

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/red-hat-storage/odf-operator
 
 go 1.22.7
-
+
```
iamniting, line 4: This change is not required.

iamniting (changes requested):

go.mod
iamniting: I still see this change.

kasserater: @iamniting very weird, i was sure i removed it
i pushed another fix now

kasserater:
@k8s-ci-robot

kasserater:
@openshift-ci-robot /retest

kasserater:
@iamniting who can assist with resolving the Jira issue?

iamniting:
> @iamniting who can assist with resolving the Jira issue?

Let the 4.18.3 release. I will get it sorted.

iamniting (approved):

PR #619, opened 2025-05-27
add odf-snapshot-controller bundle

Nikhil-Ladha:
add new odf-snapshot-controller bundle to odf-op
Ref: https://issues.redhat.com/browse/RHSTOR-7375


hack/make-bundle-vars.mk
```diff
@@ -137,6 +137,11 @@ IBM_ODF_SUBSCRIPTION_PACKAGE ?= ibm-storage-odf-operator
 IBM_ODF_SUBSCRIPTION_CHANNEL ?= stable-v1.7
 IBM_ODF_SUBSCRIPTION_CSVNAME ?= $(IBM_ODF_SUBSCRIPTION_PACKAGE).$(IBM_ODF_BUNDLE_VERSION)
 
+ODF_SNAPSHOT_CONTROLLER_BUNDLE_IMG ?= quay.io/ocs-dev/snapshot-controller-bundle:latest
```
iamniting, line 140: Lets not use the latest tag, It creates problems when the new image is pushed on the same tag.

iamniting, line 140: Lets move this after csi addons

iamniting (changes requested):
Can you pls separate the generated changes and manual changes?

config/manager/configmap.yaml
```diff
@@ -60,3 +60,7 @@ data:
     pkg: rook-ceph-operator
     scaleUpOnInstanceOf:
       - cephclusters.ceph.rook.io
+  SNAPSHOT_CONTROLLER: |
```
iamniting, line 63: This is also a generated change.

iamniting (changes requested):

config/manager/configmap.yaml
Nikhil-Ladha: Updated.

hack/make-files.mk
```diff
@@ -18,6 +18,10 @@ data:
     pkg: $(CSIADDONS_SUBSCRIPTION_PACKAGE)
     scaleUpOnInstanceOf:
       - cephclusters.ceph.rook.io
+  SNAPSHOT_CONTROLLER: |
+    channel: $(ODF_SNAPSHOT_CONTROLLER_SUBSCRIPTION_CHANNEL)
+    csv: $(ODF_SNAPSHOT_CONTROLLER_SUBSCRIPTION_CSVNAME)
+    pkg: $(ODF_SNAPSHOT_CONTROLLER_SUBSCRIPTION_PACKAGE)
```
iamniting, line 24: All other things looks good to me, Now the only thing left is when should we scale this up, on presence of cephcluster or storagecluster or noobaa?

iamniting (changes requested):

iamniting (approved):

PR #647, opened 2025-07-17
generate a cnsa-dependencies bundle

iamniting:
This PR helps in generating a cnsa-dependencies bundle which will be used as a parent pkg to deploy the IBM spectrum scale operator and lock it to a particular version. 


leelavg (commented):
just so I'm understanding it correctly, this entire PR is for us to test if we are able to manage deployments across multiple namespaces, isn't it?

As per the design we would just need indication or config to start deploying CNSA, isn't it?

iamniting:
> just so I'm understanding it correctly, this entire PR is for us to test if we are able to manage deployments across multiple namespaces, isn't it?
> 
> As per the design we would just need indication or config to start deploying CNSA, isn't it?

This PR does not do any of the controller work, We are just creating a new bundle in this PR.

PR #670, opened 2025-08-18
webhook: mutate rook op csv if previeous csv had hostNetwork enabled

rohan47:
updated webhook to mutate csv if the previous csv had hostNetwork enabled for deployment in order to persist csv update from https://github.com/red-hat-storage/ocs-operator/pull/3423

Bug: https://issues.redhat.com/browse/DFBUGS-3832

### DIFF
--- a/webhook/csv.go
+++ b/webhook/csv.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/go-logr/logr"
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


rohan47:
> > We should not write rook csv specific code, We should have a generic code which works for all csvs.
> 
> Generic code will not work in this case as we need t update rook-ceph-operator deployment specifically, if we want generic then the configmap needs to be updated

Making changes as dicsussed

iamniting (approved):
/lgtm @nb-ohad Can you take a look

webhook/csv.go
```diff
@@ -123,35 +137,35 @@ func (r *ClusterServiceVersionDeploymentScaler) loadOdfConfigMapData(ctx context
...
+	for i := range prevDeployments {
+		prevDeployment := &prevDeployments[i]
 
-	if err := r.Client.Get(ctx, key, prevCsv); errors.IsNotFound(err) {
-		// new install where an previous csv does not exists
-		return false, nil
-	} else if err != nil {
-		logger.Error(err, "failed getting previous csv")
-		return false, err
+		for j := range newDeployments {
```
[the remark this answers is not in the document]
iamniting: @nb-ohad We are doing that only without creating any maps and all. This logic is much simpler. Also, I am not worried about the complexity here because we will have only a few deployments.

webhook/csv.go
```diff
@@ -71,19 +74,33 @@ func (r *ClusterServiceVersionDeploymentScaler) Handle(ctx context.Context, req
...
+		key := client.ObjectKey{Name: csv.Spec.Replaces, Namespace: csv.Namespace}
+		if err := r.Client.Get(ctx, key, prevCsv); err != nil {
+			if client.IgnoreNotFound(err) != nil {
+				logger.Error(err, "failed to get previous CSV", "csv", csv.Spec.Replaces)
+				return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed getting previous csv %q: %w", csv.Spec.Replaces, err))
+			}
+			noPrevCsvFound = true
+		} else {
+			logger.Info("previous CSV found", "csv", csv.Spec.Replaces)
+			defaultsMap := map[string]int32{}
```
iamniting, line 89: I would prefer a type of value to be `any` if possible.

webhook/csv.go
```diff
@@ -71,19 +74,33 @@ func (r *ClusterServiceVersionDeploymentScaler) Handle(ctx context.Context, req
...
+				defaultsMap[keyReplicas] = 0
+			}
+			r.syncNewCsvWithPrevCsv(prevCsv, csv, defaultsMap)
+		}
 	}
 
-	if running {
-		logger.Info("ignoring csv as the previous csv deployments are running")
-		return admission.Allowed("previous csv deployments are running")
+	if noPrevCsvFound {
```
iamniting, line 99: This logic is also broken as in the case of a new install, `noPrevCsvFound` will be false (zero value) and it will never scale down. This will be automatically solved if you use the above variable and flip the conditions.

iamniting (changes requested):

webhook/csv.go
```diff
@@ -71,19 +75,35 @@ func (r *ClusterServiceVersionDeploymentScaler) Handle(ctx context.Context, req
...
+		if err := r.Client.Get(ctx, key, prevCsv); err != nil {
+			if client.IgnoreNotFound(err) != nil {
+				logger.Error(err, "failed to get previous CSV", "csv", csv.Spec.Replaces)
+				return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed getting previous csv %q: %w", csv.Spec.Replaces, err))
+			}
+		} else {
+			logger.Info("previous CSV found", "csv", csv.Spec.Replaces)
+			isPrevCsvFound = true
+			defaultsMap := map[string]any{}
+			// in case previours csv has n deployments and current csv has n+1 deployments the the defaults map will be used to scale down replicas of the new deployment
```
iamniting:
```suggestion
			// in case previours csv has n deployments and current csv has n+1 deployments the the defaults map will be used to set the default value of replicas for the new deployment
```

iamniting (changes requested):
Can you also change the commit msg to include the DNSPolicy?

iamniting (approved):

PR #679, opened 2025-09-08
controllers: update catalog source for all dependencies subscriptions

iamniting:
Previously, we only updated the catalog source for the odf-dependencies subscription. This change extends the logic to update the catalog source for odf-dependencies as well as cnsa-dependencies subscription.

### DIFF
--- a/controllers/subscriptions.go
+++ b/controllers/subscriptions.go
@@ -102,10 +102,12 @@ func GetDesiredSubscription(ctx context.Context, cli client.Client, record *OlmP
 	}
 
 	if !subExsist {
-		// Set the catalog source for the odf-dependencies subscription to match that of the odf-operator subscription
-		// This ensures that the odf-dependencies subscription uses the same catalog source across all environments,
+		// Set the catalog source for the dependencies subscription to match that of the odf-operator subscription
+		// This ensures that the dependencies subscription uses the same catalog source across all environments,
 		// including offline and test environments where the catalog name may vary.
-		if desiredSubscription.Spec.Package == OdfDepsSubscriptionPackage {
+		if desiredSubscription.Spec.Package == OdfDepsSubscriptionPackage ||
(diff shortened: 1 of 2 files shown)


controllers/subscriptions.go
```diff
@@ -102,10 +102,10 @@ func GetDesiredSubscription(ctx context.Context, cli client.Client, record *OlmP
 	}
 
 	if !subExsist {
-		// Set the catalog source for the odf-dependencies subscription to match that of the odf-operator subscription
-		// This ensures that the odf-dependencies subscription uses the same catalog source across all environments,
+		// Set the catalog source for the dependencies subscription to match that of the odf-operator subscription
+		// This ensures that the dependencies subscription uses the same catalog source across all environments,
 		// including offline and test environments where the catalog name may vary.
-		if desiredSubscription.Spec.Package == OdfDepsSubscriptionPackage {
+		if strings.HasSuffix(desiredSubscription.Spec.Package, "-dependencies") {
```
leelavg, line 108: don't we know the name of all packages that we are going to manage during the build time itself?

better to change to strings.Contain for more clarity at the expense of little overhead.

iamniting: Yes we do know the package name, But we wanted to make a change in such a way that it wont require changes in the future. Even earlier changes are also like this. This was missed from the earlier change. https://github.com/red-hat-storage/odf-operator/pull/652

leelavg: ack, then I suggest to change in places where we did this mistake. If we didn't know the full name, prefix or suffix would've been good, that's not the case here.

iamniting: But this will leave us to change it whenever we introduce a new dependency. Which we wanted to avoid.

iamniting: I am using both the names now, Pls take a look.

leelavg (approved):

PR #694, opened 2025-10-16
controllers: Make operator not upgradeable if ODF minor version is ahead of OCP

malayparida2000:
The supportability matrix allows ODF minor version to be upto n+1 ahead of OCP minor version (i.e ODF X.Y+1, OCP X.Y). If the ODF minor version is already ahead of OCP minor version (e.g. ODF 4.21.Z, OCP 4.20.Z), further ODF upgrade would make it incompatible with the current OCP version, hence mark the operator as not upgradeable.

To resolve this user would have to either upgrade OCP to the next minor version or override the operator upgradeable condition via the operatorCondition CR.
Ref-https://issues.redhat.com/browse/RHSTOR-8012

### DIFF
--- a/controllers/subscription_controller.go
+++ b/controllers/subscription_controller.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/blang/semver/v4"
 	"github.com/red-hat-storage/odf-operator/pkg/util"
 )
 
@@ -286,6 +287,16 @@ func (r *SubscriptionReconciler) ensureSubscriptions(ctx context.Context, logger
 }
 
 func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logger logr.Logger, condMap map[string]struct{}) error {
+	// Make operator not upgradeable if ODF minor version is ahead of OCP minor version(e.g. ODF 4.21.z, OCP 4.20.z)
+	if isODFAhead, err := r.isODFAheadOfOCP(ctx); err != nil {
+		return err
+	} else if isODFAhead {
+		logger.Info("ODF minor version ahead of OCP. Further upgrade would reach an unsupported config, marking the operator as not upgradeable")
+		return r.operatorCondition.Set(ctx, metav1.ConditionFalse,
+			conditions.WithReason("ODFVersionAheadOfOCP"),
+			conditions.WithMessage("ODF version is already ahead of OCP. Further ODF upgrade would make it incompatible with the current OCP version"))
+	}
+
 	ocdList := &opv2.OperatorConditionList{}
 	err := r.Client.List(ctx, ocdList, client.InNamespace(r.OperatorNamespace))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/subscription_controller.go
```diff
@@ -310,6 +312,18 @@ func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logge
 		}
 	}
 
+	// Make operator not upgradeable if ODF version is n+1 ahead of OCP version
```
leelavg: at multiple places I see `n`, I suppose `n` is referring to OCP, better be explicit, if OCP is at 4.Y and ODF is at 4.(Y+1) do not allow ODF to upgrade further.

controllers/subscription_controller.go
```diff
@@ -310,6 +312,18 @@ func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logge
...
+	isODFAhead, err := isODFAheadOfOCP(ctx, r.Client, r.operatorConditionName)
+	if err != nil {
+		return err
+	}
+	if isODFAhead {
+		msg := "ODF version is already ahead of OCP. Further ODF upgrade would make it incompatible with the current OCP version"
```
leelavg: And why aren't we enforcing ODF version resulting in ODF being already ahead of OCP?

controllers/subscription_controller.go
```diff
@@ -433,3 +447,25 @@ func getNotUpgradeableCond(ocd *opv2.OperatorCondition) *metav1.Condition {
...
+		return false, fmt.Errorf("failed to determine OCP version: %v", err)
+	}
+
+	ocpVersion, err := semver.Make(ocpVersionString)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse OCP version %s: %v", ocpVersionString, err)
+	}
+
+	// Parse ODF version from operatorConditionName as it's the same as the CSV name
+	version := operatorConditionName[strings.Index(operatorConditionName, ".v")+2:]
```
leelavg: better not, let's either take this as the actual csv name and get the version from it or traverse through the Pod -> Deployment -> CSV (iirc, being done in odf-op already at some place) to get the version.

I know this is easier but there are some variations in the csv name, sometimes it has `stable`, `rhodf` and just empty depending on environment and mostly doing name mangling on not established/not controlled entity (by us) is problematic.

controllers/subscription_controller.go
```diff
@@ -433,3 +447,25 @@ func getNotUpgradeableCond(ocd *opv2.OperatorCondition) *metav1.Condition {
...
+	odfVersion, err := semver.Make(version)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse ODF version from operatorConditionName %s: %v", operatorConditionName, err)
+	}
+
+	return odfVersion.Major != ocpVersion.Major || odfVersion.Minor > ocpVersion.Minor, nil
```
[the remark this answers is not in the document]
malayparida2000: >ODF will not upgrade from 4.20 to 4.22 in a single step

Wasn't work for enabling EUS upgrade based on making this possible?

controllers/subscription_controller.go
malayparida2000: As per offline discussion now updated to get the csv with the name & then get the version from it.

PR #700, opened 2025-11-19
controllers: skip scaling operators if Fusion-managed Cluster CR exists

iamniting:
Add logic to avoid scaling CNSA operators in the ibm-spectrum-scale namespace when a Fusion-managed Cluster CR is present. This prevents multiple CNSA operator instances from running simultaneously.


iamniting:
This PR is not required anymore

PR #719, opened 2026-01-06
controllers: add resource requirement for ux-backend-server deployment

rohan47:
resource requirement was missing from ux-backend-server deployment


Nikhil-Ladha (approved):
/lgtm
/approve

PR #723, opened 2026-01-16
Move ux backend-server code from ocs-operator to odf-operator

rohan47:
ux-backend-server deployment was handled by odf-operator while the code was in ocs-operator. With this PR the code for ux-backend-server, RBACs and build is moved to odf-operator

### DIFF
--- a/bundle/odf-operator/manifests/oauth-proxy-role-binding_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
+++ b/bundle/odf-operator/manifests/oauth-proxy-role-binding_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  creationTimestamp: null
+  name: oauth-proxy-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: oauth-proxy-role
+subjects:
+- kind: ServiceAccount
+  name: ux-backend-server
+  namespace: openshift-storage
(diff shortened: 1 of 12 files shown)


hack/make-bundle-vars.mk
```diff
@@ -45,6 +45,7 @@ IMAGE_REGISTRY ?= quay.io
 REGISTRY_NAMESPACE ?= ocs-dev
 IMAGE_TAG ?= latest
 IMAGE_NAME ?= odf-operator
+DEVICEFINDER_IMAGE_NAME ?= devicefinder
```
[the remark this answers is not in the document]
rohan47: updated

controllers/uxbackend.go
```diff
@@ -56,7 +56,7 @@ func getUXBackendServerDeployment() *appsv1.Deployment {
 						},
 						Image:           os.Getenv("UX_BACKEND_SERVER_IMAGE"),
 						ImagePullPolicy: "IfNotPresent",
-						Command:         []string{"/usr/local/bin/ux-backend-server"},
+						Command:         []string{"/ux-backend-server"},
```
iamniting: Can you make sure the path is same even in the DS?

Makefile
```diff
@@ -108,6 +108,14 @@ docker-build: godeps-update test-setup go-test ## Build docker image with the ma
 docker-push: ## Push docker image with the manager.
 	docker push ${IMG}
 
+devicefinder-build: ## Build devicefinder binary.
+	pushd services/devicefinder
+	docker build -f Dockerfile -t "${DEVICEFINDER_IMAGE}" . --build-arg="LDFLAGS=${LDFLAGS}"
```
iamniting: we dont need to give the file name as it will automatically picks up the `Dockerfile`

go.mod
```diff
@@ -1,106 +1,171 @@
 module github.com/red-hat-storage/odf-operator
 
-go 1.24.3
+go 1.24.6
+
+toolchain go1.24.10
```
leelavg: better have both go and toolchain point to the same version, pick whatever is available, when toolchain > go, toolchain version is explicitly downloaded (if higher/same doesn't exist) in users machine to build this project

controllers/uxbackend.go
rohan47: DS has a different path for manager binary as well, The current path in the PR is same as manager binary

b-ranto: We used the same path as before when enabling this, i.e. `/usr/local/bin/ux-backend-server`, the changes are not in, it can be changed if desired.

rohan47: Can you please change it to `/ux-backend-server`

PR #774, opened 2026-04-22
deps: add ocs-tls-profiles package as one of the olm dependencies

leelavg:
ocs-tls-profiles is a new package with only CRD and CSV without any deployment which'll be used by ocs-operator for configuring TLS settings on DF products and be installed as part of odf-dependencies.


leelavg:
Depends on https://github.com/red-hat-storage/ocs-operator/pull/3767

catalog/odf-dependencies.yaml
```diff
@@ -44,7 +44,7 @@ properties:
 - type: olm.package.required
   value:
     packageName: rook-ceph-operator
-    versionRange: 4.22.0
+    versionRange: 4.20.0
```
leelavg: not exactly sure why the versionRange is changed to 4.20.0 here, any hints here @iamniting thanks!

iamniting: Even i am not sure, why this happen, Can you try to rebase?

leelavg: the branch https://github.com/leelavg/odf-operator/tree/tls-pkg is already on top of main, i'll relook as the CI is also failing.

leelavg: locally `make catalog` pulled the deps bundle image existing in quay which had 4.20 as deps and in GH CI it actually builds again with correctly defined versions.

iamniting (approved):

PR #775, opened 2026-04-23
controllers: handle non-default replica counts for specific deployments

iamniting:
Ensure all CSV deployments have a valid replica count by defaulting to 1. Introduce a configurable map to override this default for specific deployments that require higher replica counts.

This prevents under-scaling of deployments that are expected to run with more than one replica.

https://redhat.atlassian.net/browse/DFBUGS-6018


Nikhil-Ladha (approved):
/lgtm

PR #795, opened 2026-05-12
controllers: block upgrade when CSVs are on non-desired version

iamniting:

### DIFF
--- a/controllers/subscription_controller.go
+++ b/controllers/subscription_controller.go
@@ -101,7 +101,7 @@ func (r *SubscriptionReconciler) Reconcile(ctx context.Context, _ ctrl.Request) 
 		return ctrl.Result{}, err
 	}
 
-	if err := r.setOperatorCondition(ctx, logger, csvNamesMap); err != nil {
+	if err := r.setOperatorCondition(ctx, logger, csvNamesMap, olmPkgRecords); err != nil {
 		return ctrl.Result{}, err
 	}
 
@@ -287,7 +287,7 @@ func (r *SubscriptionReconciler) ensureSubscriptions(ctx context.Context, logger
 	return combinedErr
 }
 
-func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logger logr.Logger, condMap map[string]struct{}) error {
+func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logger logr.Logger, condMap map[string]struct{}, olmPkgRecords []*OlmPkgRecord) error {
 	// Make operator not upgradeable if ODF minor version is ahead of OCP minor version(e.g. ODF 4.21.z, OCP 4.20.z)
 	if isODFAhead, err := r.isODFAheadOfOCP(ctx); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


iamniting:
@Nikhil-Ladha We need to test this thoroughly. I’m concerned about z-stream upgrades, where ODF may block its own upgrade while other operators continue upgrading independently. In that case, odf-operator could get stuck waiting for the expected child CSVs, while those CSVs may have already moved to a newer version.

controllers/subscription_controller.go
```diff
@@ -308,6 +308,35 @@ func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logge
 			conditions.WithMessage("OCP upgrade is incomplete. ODF is not upgradeable to ensure cluster stability"))
 	}
 
+	for _, olmPkgRecord := range olmPkgRecords {
+
+		csvList := &opv1a1.ClusterServiceVersionList{}
+		labelKey := fmt.Sprintf("operators.coreos.com/%s.%s", olmPkgRecord.Pkg, olmPkgRecord.Namespace)
+		err := r.Client.List(ctx, csvList, client.InNamespace(olmPkgRecord.Namespace), client.HasLabels{labelKey})
```
leelavg, line 315: since all csvs form a set in a namespace could we do a list once and then filter? I'm worried that by listing individually we'd miss some updates and rather process all csvs as a set at that specific instant?

I'm thinking like map[string]*csv where string is the label and can pull the csv based on olmpkgrecord interpolation, any other way is also fine.

controllers/subscription_controller.go
```diff
@@ -308,6 +308,35 @@ func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logge
...
+			logger.Error(err, "failed to list ClusterServiceVersion")
+			return err
+		}
+
+		if len(csvList.Items) == 0 {
+			continue
+		} else if len(csvList.Items) == 1 {
+			if csvList.Items[0].Name != olmPkgRecord.Csv {
+				logger.Info("Desired csv version not found, marking the operator as not upgradeable",
+					"desiredCsv", olmPkgRecord.Csv, "actualCsv", csvList.Items[0].Name)
```
leelavg, line 326: there is a chance with this odf-op can never get upgraded if dependent already got upgraded for a patch version which could also stop next minor upgrade?

should we also make use of replaces field in the csv?

controllers/subscription_controller.go
```diff
@@ -308,6 +308,35 @@ func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logge
...
+				return r.operatorCondition.Set(ctx, metav1.ConditionFalse,
+					conditions.WithReason("DesiredCsvVersionNotFound"),
+					conditions.WithMessage(fmt.Sprintf("Expecting CSV version %s, But found %s", olmPkgRecord.Csv, csvList.Items[0].Name)))
+			}
+		} else if len(csvList.Items) > 1 {
+			logger.Info("More than one csvs found in the cluster, marking the operator as not upgradeable",
+				"pkgName", olmPkgRecord.Pkg)
+			return r.operatorCondition.Set(ctx, metav1.ConditionFalse,
```
leelavg, line 334: lg, we are in the process of upgrade and stop odf-op from moving forward further?

controllers/subscription_controller.go
```diff
@@ -308,6 +308,35 @@ func (r *SubscriptionReconciler) setOperatorCondition(ctx context.Context, logge
...
+				conditions.WithReason("DesiredCsvVersionNotFound"),
+				conditions.WithMessage(fmt.Sprintf("The Csvs are being upgraded for pkg %s", olmPkgRecord.Pkg)))
+		}
+	}
```
leelavg, line 338: if backported to .z then do we need some extra initialization code expecting that dependents are already ahead of odf-op?

thinking about some combinations:

---

t0: everything at 4.21.1
t1: odf-op at 4.21.1 but recipe at 4.21.2 <- no new code, odf-op gets upgraded, good.
t2: everything at 4.21.2
t3: odf-op at 4.21.2 and recipe at 4.21.3, 4.21.2 (in progress) <- odf-op marks unupgradeable (3rd cond)
t4: odf-op at 4.21.2 and recipe at 4.21.3 <- odf-op unupgradeable (2nd cond)

now t4 is deadlocked? took recipe as an example as it doesn't need to pull any image.

---

t0: everything at 4.21.9
t1: odf-op inprogress to 4.22.0 and dependents are at 4.21.9
t2: odf-op at 4.22.0 and ocs-op is at 4.21.9, 4.22.0 (in progress) <- odf-op marks unupgradeable (3rd cond)
t3: odf-op at 4.22.0 and ocs-op is at 4.22.0 <- odf-op upgradeable (out of the loop)

all good. took ocs-op as an example as it needs to pull an image by which time odf-op is already running newer version

t4: odf-op and deps races for 4.22.1 <- possible odf-op can mark unupgradeable (2nd cond when older odf-op with current code looks at upgraded recipe)

now t4 is deadlocked.

---

leelavg (approved):

iamniting:
It is very tricky to solve all problem and we are also moving towards olm v1. It is not worth it to fix the problem in olm v0 based odf operator. Hence i am closing it.

PR #797, opened 2026-05-15
edit odf subscription and add annotation

AnikaYadav:
edit ODF Subscription and add annotation ([RHSTOR-8926](https://redhat.atlassian.net/browse/RHSTOR-8926))
Includes two subtasks:
1- add annotation to odf-operator CSV to claim support for GCP STS (RHSTOR 8932)
2- add envs to noobaa-operator subscription for GCP STS (RHSTOR 8933)

### DIFF
--- a/controllers/subscriptions.go
+++ b/controllers/subscriptions.go
@@ -371,11 +371,15 @@ func AdjustSpecialCasesSubscriptionConfig(subscription *opv1a1.Subscription) {
 
 	case "noobaa-operator", "mcg-operator":
 		var envVars []corev1.EnvVar
-		for _, name := range []string{"ROLEARN", "CLIENTID", "TENANTID", "SUBSCRIPTIONID", "RESOURCEGROUP"} {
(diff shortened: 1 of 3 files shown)


controllers/subscriptions.go
```diff
@@ -376,6 +376,14 @@ func AdjustSpecialCasesSubscriptionConfig(subscription *opv1a1.Subscription) {
 				envVars = append(envVars, corev1.EnvVar{Name: name, Value: value})
 			}
 		}
+		if subscription.Spec.Package == "noobaa-operator" {
+			for _, name := range []string{"PROJECT_NUMBER", "POOL_ID", "PROVIDER_ID", "SERVICE_ACCOUNT_EMAIL"} {
```
iamniting: Can you add these variables to the above list. This case is for noobaa itself.

iamniting (changes requested):

iamniting (approved):

PR #847, opened 2026-06-29
Updated pods to have annotations with openshift.io/required-scc

ezio-auditore:
Epic: [RHSTOR-8757](https://redhat.atlassian.net/browse/RHSTOR-8757) [Enforce least-privileged Security Context Constraint (SCC) across ODF pods ](https://redhat.atlassian.net/browse/RHSTOR-8757)
Summary
Pin every pod template owned by odf-operator to an explicit Security Context Constraint using the openshift.io/required-scc annotation. This makes SCC assignment deterministic for audit and compliance, instead of relying on OpenShift admission to pick the most permissive matching SCC at runtime.

Changes cover all four workloads managed by this operator:

- odf-operator-controller-manager — restricted-v2 (config/manager/manager.yaml, bundle CSV)
- odf-console — restricted-v2 (config/console/console_init.yaml, bundle CSV)
- ux-backend-server — restricted-v2 (controllers/uxbackend.go, runtime Deployment)
- devicefinder-discovery — privileged (services/devicefinder/assets/template.go, runtime DaemonSet; requires host /dev access and a privileged container)
- devicefinder-discovery keeps its existing target.workload.openshift.io/management annotation; the privileged SCC annotation is added alongside it.

### DIFF
--- a/controllers/uxbackend.go
+++ b/controllers/uxbackend.go
@@ -51,6 +51,9 @@ func getUXBackendServerDeployment(tolerations []corev1.Toleration) *appsv1.Deplo
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: labels,
+				Annotations: map[string]string{
+					"openshift.io/required-scc": "restricted-v2",
+				},
 			},
 			Spec: corev1.PodSpec{
 				Containers: []corev1.Container{
--- a/services/devicefinder/assets/template.go
+++ b/services/devicefinder/assets/template.go
@@ -30,6 +30,7 @@ func NewDeviceFinderDaemonSet(namespace string, containerImage string) *appsv1.D
 					},
 					Annotations: map[string]string{
 						"target.workload.openshift.io/management": `{"effect": "PreferredDuringScheduling"}`,
+						"openshift.io/required-scc":               "privileged",
 					},
 				},
 				Spec: corev1.PodSpec{
--- a/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
@@ -405,6 +405,7 @@ spec:
             metadata:
               annotations:
                 kubectl.kubernetes.io/default-container: manager
+                openshift.io/required-scc: restricted-v2
               labels:
                 app.kubernetes.io/name: odf-operator
                 control-plane: controller-manager
@@ -502,6 +503,8 @@ spec:
           strategy: {}
           template:
             metadata:
+              annotations:
+                openshift.io/required-scc: restricted-v2
               labels:
                 app: odf-console
             spec:
--- a/config/console/console_init.yaml
+++ b/config/console/console_init.yaml
@@ -9,6 +9,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        openshift.io/required-scc: restricted-v2
       labels:
         app: odf-console
     spec:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -23,6 +23,7 @@ spec:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
+        openshift.io/required-scc: restricted-v2
       labels:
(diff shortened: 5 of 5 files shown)


services/devicefinder/assets/template.go
```diff
@@ -30,6 +30,7 @@ func NewDeviceFinderDaemonSet(namespace string, containerImage string) *appsv1.D
 					},
 					Annotations: map[string]string{
 						"target.workload.openshift.io/management": `{"effect": "PreferredDuringScheduling"}`,
+						"openshift.io/required-scc":               "privileged",
```
rohan47, line 33: If I understand correctly, this annotation is used just for identifying the scc used by the pod.
And the actual scc permissions come from [cluster-role](https://github.com/red-hat-storage/odf-operator/blob/main/config/ux-backend/devicefinder-cluster-role.yaml) right?

controllers/uxbackend.go
```diff
@@ -51,6 +51,9 @@ func getUXBackendServerDeployment(tolerations []corev1.Toleration) *appsv1.Deplo
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: labels,
+				Annotations: map[string]string{
+					"openshift.io/required-scc": "restricted-v2",
+				},
```
iamniting, line 56: Just adding annotation here wont add the annotation on the existing/upgraded clusters. Can you also add the annotations logic in the `CreateOrUpdate` func where this func is being called.

services/devicefinder/assets/template.go
iamniting, line 33: Just adding annotation here wont add the annotation on the existing/upgraded clusters. Can you also add the annotations logic in the CreateOrUpdate func where this func is being called.

iamniting (changes requested):

controllers/uxbackend.go
ezio-auditore: But, in the exisitng code all label and annotations ar parts of the `getUXBackendServerDeployment` func.
Should I also explicity add under CreateOrUpdate call?

services/devicefinder/assets/template.go
ezio-auditore: same for this, a previous annotation also exists , does that mean `"target.workload.openshift.io/management": `{"effect": "PreferredDuringScheduling"}`` is not applied on update?
@rohan47 do you know about this?

controllers/uxbackend.go
iamniting: Yes

services/devicefinder/assets/template.go
ezio-auditore: yes

controllers/uxbackend.go
ezio-auditore: Updated

services/devicefinder/assets/template.go
ezio-auditore: Updated this as well

controllers/uxbackend.go
iamniting: @ezio-auditore I overlooked these changes earlier. They aren't required because they're part of the spec. If they had been part of the metadata, the changes would have been necessary. Could you please revert these changes?

services/devicefinder/assets/template.go
iamniting: @ezio-auditore I overlooked these changes earlier. They aren't required because they're part of the spec. If they had been part of the metadata, the changes would have been necessary. Could you please revert these changes?

ezio-auditore: ok, done

iamniting: I still see those changes.

controllers/uxbackend.go
iamniting: I still see those changes.

ezio-auditore: fixed

services/devicefinder/assets/template.go
ezio-auditore: fixed

iamniting (approved):

PR #848, opened 2026-06-30
DFBUGS-6091: [release-4.19] switch to openshift-sustaining/grpc-go v1.75.1-sec.1 for CVE-2026-33186

malayparida2000:
Upgrading `google.golang.org/grpc` to a patched upstream release would require a Go version bump on this branch. This PR uses the OpenShift sustaining fork to backport the security fix while maintaining compatibility with the existing Go toolchain.

Ref-
https://redhat.atlassian.net/browse/DFBUGS-6091


go.mod
```diff
@@ -21,8 +21,10 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require go.opentelemetry.io/auto/sdk v1.1.0 // indirect
```
iamniting: Can you add this in the existing block?

iamniting (changes requested):

go.mod
malayparida2000: Done

iamniting (approved):

PR #868, opened 2026-07-09
controllers: fix ConsolePlugin proxy entries overwrite in ensureConsolePlugin

### LINKED ISSUE #867: ConsolePlugin proxy entries flapping: odf-operator overwrites proxy entries added by ocs-operator
## Description

The `odf-console` ConsolePlugin's `spec.proxy` entries are continuously toggling — the `internalRgwS3` proxy entry (pointing to `rook-ceph-rgw-ocs-storagecluster-cephobjectstore`) is added by the ocs-operator and then immediately removed by the odf-operator, in an infinite loop.

This causes unnecessary API churn (the ConsolePlugin `generation` climbs into the tens of thousands) and downstream reconcile noise in the console-operator.

## Root Cause

Two controllers both manage `spec.proxy` on the `odf-console` ConsolePlugin using full-object `Update` calls, and each overwrites the other's entries:

1. **ocs-operator** (`console_configuration.go`) reads the ConsolePlugin, checks if the `internalRgwS3` alias is present, and appends it if missing.
2. **odf-operator** (`clusterversion_controller.go:180`) calls `controllerutil.CreateOrUpdate` with a mutate function that **unconditionally replaces** `Spec.Proxy` with a hard-coded list from `GetConsolePluginProxy()` (`console/console.go:84-146`). This list does not include `internalRgwS3`, so the entry is removed every reconcile.

The smoking gun is `clusterversion_controller.go` line 180:
```go
odfConsolePlugin.Spec.Proxy = console.GetConsolePluginProxy(OperatorNamespace)
```

This creates an infinite loop:
1. ocs-operator adds `internalRgwS3` → ConsolePlugin now has N+1 proxy entries
2. odf-operator reconciles → replaces proxy list with N entries (no `internalRgwS3`)
3. ocs-operator sees `internalRgwS3` is missing → adds it back → goto 1

## Evidence

`managedFields` on the ConsolePlugin confirms the two writers:
```
Manager: manager       Fields: spec.backend, spec.displayName, spec.i18n
Manager: ocs-operator  Fields: spec.proxy
```

Watching the ConsolePlugin shows the `rook-ceph-rgw-ocs-storagecluster-cephobjectstore` endpoint toggling on/off every few seconds:
```
$ oc get consoleplugin odf-console -o jsonpath='{.spec.proxy[*].endpoint}{"\n"}' -w
# alternates between 4 and 5 proxy entries continuously
```

## Suggested Fix

In `ensureConsolePlugin`, instead of unconditionally replacing the proxy list, preserve any proxy entries not owned by the odf-operator. For example, merge the known entries with any unknown ones already present on the resource.

ygalblum:
## Summary
- `ensureConsolePlugin` unconditionally replaced the entire `spec.proxy` list on the `odf-console` ConsolePlugin every reconcile, discarding proxy entries added by other controllers (e.g. `internalRgwS3` from ocs-operator)
- This caused an infinite reconcile loop where ocs-operator added the entry and odf-operator immediately removed it, generating tens of thousands of unnecessary API writes
- Added `MergeConsolePluginProxy()` that updates known entries in place while preserving unknown entries from other controllers

## Root Cause

In `clusterversion_controller.go:180`, the mutate function inside `controllerutil.CreateOrUpdate` did:
```go
odfConsolePlugin.Spec.Proxy = console.GetConsolePluginProxy(OperatorNamespace)
```
This replaced the full proxy slice with a hard-coded 5-entry list, removing the `internalRgwS3` entry that `ocs-operator` appends. Since `CreateOrUpdate` uses a full-object `Update` (not patch/SSA), the removal was written to the API server, triggering the ocs-operator to add it back.

## Fix

Replace the full-list assignment with a merge operation:
```go
odfConsolePlugin.Spec.Proxy = console.MergeConsolePluginProxy(odfConsolePlugin.Spec.Proxy, OperatorNamespace)
```
`MergeConsolePluginProxy` iterates the existing proxy list, updates entries with matching aliases to their desired state, and preserves entries with unknown aliases.

## Test plan
- [x] Unit tests for `MergeConsolePluginProxy` covering: preserving unknown entries, updating known entries, empty existing list
- [x] `go build ./...` passes
- [x] `go test ./console/` passes

Fixes: https://github.com/red-hat-storage/odf-operator/issues/867

### DIFF
--- a/console/console.go
+++ b/console/console.go
@@ -146,6 +146,35 @@ func GetConsolePluginProxy(serviceNamespace string) []consolev1.ConsolePluginPro
 	}
 }
 
+// MergeConsolePluginProxy merges the desired proxy entries into the existing
+// list, preserving any entries added by other controllers (e.g. ocs-operator).
+// Entries owned by this operator (matched by alias) are updated in place;
+// unknown entries are kept as-is.
+func MergeConsolePluginProxy(existing []consolev1.ConsolePluginProxy, serviceNamespace string, removeAliases []string) []consolev1.ConsolePluginProxy {
+	aliasesToRemove := make(map[string]bool, len(removeAliases))
+	for _, alias := range removeAliases {
+		aliasesToRemove[alias] = true
+	}
+
+	desired := GetConsolePluginProxy(serviceNamespace)
+
+	desiredAliases := make(map[string]bool, len(desired))
+	for _, proxy := range desired {
+		desiredAliases[proxy.Alias] = true
+	}
+
+	merged := append([]consolev1.ConsolePluginProxy{}, desired...)
+
+	for _, proxy := range existing {
+		if aliasesToRemove[proxy.Alias] || desiredAliases[proxy.Alias] {
+			continue
+		}
+		merged = append(merged, proxy)
+	}
+
+	return merged
+}
+
 func GetConsolePluginCR(consolePort int32, serviceNamespace string) *consolev1.ConsolePlugin {
 	return &consolev1.ConsolePlugin{
 		ObjectMeta: metav1.ObjectMeta{
--- a/console/console_test.go
+++ b/console/console_test.go
@@ -1,0 +1,113 @@
+package console
+
+import (
+	"testing"
+
+	consolev1 "github.com/openshift/api/console/v1"
+)
+
+func TestMergeConsolePluginProxy_PreservesUnknownEntries(t *testing.T) {
+	ns := "openshift-storage"
+	extra := consolev1.ConsolePluginProxy{
+		Alias: "internalRgwS3",
+		Endpoint: consolev1.ConsolePluginProxyEndpoint{
+			Type: consolev1.ProxyTypeService,
+			Service: &consolev1.ConsolePluginProxyServiceConfig{
+				Name:      "rook-ceph-rgw-ocs-storagecluster-cephobjectstore",
+				Namespace: ns,
+				Port:      443,
+			},
+		},
+		Authorization: consolev1.None,
+	}
+
+	existing := append(GetConsolePluginProxy(ns), extra)
+	merged := MergeConsolePluginProxy(existing, ns, nil)
+
+	found := false
+	for _, p := range merged {
+		if p.Alias == "internalRgwS3" {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Error("MergeConsolePluginProxy dropped the internalRgwS3 entry added by another controller")
+	}
+
+	desired := GetConsolePluginProxy(ns)
+	if len(merged) != len(desired)+1 {
+		t.Errorf("expected %d proxy entries, got %d", len(desired)+1, len(merged))
+	}
+}
+
+func TestMergeConsolePluginProxy_UpdatesKnownEntries(t *testing.T) {
+	ns := "openshift-storage"
+	stale := []consolev1.ConsolePluginProxy{
+		{
+			Alias: "provider-proxy",
+			Endpoint: consolev1.ConsolePluginProxyEndpoint{
+				Type: consolev1.ProxyTypeService,
+				Service: &consolev1.ConsolePluginProxyServiceConfig{
+					Name:      "old-name",
+					Namespace: ns,
+					Port:      1234,
+				},
+			},
+			Authorization: consolev1.None,
+		},
+	}
+
+	merged := MergeConsolePluginProxy(stale, ns, nil)
+
+	for _, p := range merged {
+		if p.Alias == "provider-proxy" {
+			if p.Endpoint.Service.Name != "ux-backend-proxy" || p.Endpoint.Service.Port != 8888 {
+				t.Errorf("expected provider-proxy to be updated, got service=%s port=%d",
+					p.Endpoint.Service.Name, p.Endpoint.Service.Port)
+			}
+			return
+		}
+	}
+	t.Error("provider-proxy not found in merged result")
+}
+
+func TestMergeConsolePluginProxy_EmptyExisting(t *testing.T) {
+	ns := "openshift-storage"
+	merged := MergeConsolePluginProxy(nil, ns, nil)
+	desired := GetConsolePluginProxy(ns)
+
+	if len(merged) != len(desired) {
+		t.Errorf("expected %d proxy entries from empty existing, got %d", len(desired), len(merged))
+	}
+}
+
+func TestMergeConsolePluginProxy_RemovesAliases(t *testing.T) {
+	ns := "openshift-storage"
+	extra := consolev1.ConsolePluginProxy{
+		Alias: "internalRgwS3",
+		Endpoint: consolev1.ConsolePluginProxyEndpoint{
+			Type: consolev1.ProxyTypeService,
+			Service: &consolev1.ConsolePluginProxyServiceConfig{
+				Name:      "rook-ceph-rgw-ocs-storagecluster-cephobjectstore",
+				Namespace: ns,
+				Port:      443,
+			},
+		},
+		Authorization: consolev1.None,
+	}
+
+	existing := append(GetConsolePluginProxy(ns), extra)
+	merged := MergeConsolePluginProxy(existing, ns, []string{"internalRgwS3"})
+
+	for _, p := range merged {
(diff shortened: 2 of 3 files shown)


SanjalKatiyar (changes requested):
LGTM, thanks @ygalblum for the fix. We need to backport atleast till 4.21 release.

The only concern I see is that if we ever rename or remove a "desired" proxy from the `GetConsolePluginProxy` list, then during an upgrade it will be treated as an "existing" manual entry by `MergeConsolePluginProxy`. And, it will never be removed from the `ConsolePlugin` CR, these "stale entries" could accumulate over time.

I am not sure what the right long-term fix should be cc @iamniting  @bipuladh, but I think we should atleast add a comment above `MergeConsolePluginProxy` documenting this known limitation, so that we don't forget about it.

Also, `commitlint` action is failing for the CI.

prezha:
> LGTM, thanks @ygalblum for the fix. We need to backport atleast till 4.21 release.
> 
> The only concern I see is that if we ever rename or remove a "desired" proxy from the `GetConsolePluginProxy` list, then during an upgrade it will be treated as an "existing" manual entry by `MergeConsolePluginProxy`. And, it will never be removed from the `ConsolePlugin` CR, these "stale entries" could accumulate over time.
> 
> I am not sure what the right long-term fix should be cc @iamniting @bipuladh, but I think we should atleast add a comment above `MergeConsolePluginProxy` documenting this known limitation, so that we don't forget about it.
> 
> Also, `commitlint` action is failing for the CI.

i think that's a good point - just a thought on possible approach:

for those that we'd want to remove from the "standard" (hardcoded) list, perhaps we could have a separate "blacklist" that `MergeConsolePluginProxy` or even `GetConsolePluginProxy` would the use to filter them out?

that might be preferable over expanding the `consolev1.ConsolePluginProxy` struct (defined in vendor/github.com/openshift/api/console/v1/types_console_plugin.go) with another field that would indicate if the proxy is "desired", but not sure if such change could be "breaking"

ygalblum:
@prezha Yes, that was my thought as well.
@SanjalKatiyar I've added the removal functionality in a separate commit on purpose. Once we agree on this course, I'll squash the commits and address the commit lint error

prezha:
excelent, thanks @ygalblum!

just to check, please: given that the https://github.com/openshift/console-operator/pull/1107 was merged, do we also need to sort the slice to ensure it's deterministic before returning?

console/console.go
```diff
@@ -146,6 +146,47 @@ func GetConsolePluginProxy(serviceNamespace string) []consolev1.ConsolePluginPro
...
+		} else {
+			merged = append(merged, p)
+		}
+	}
+
+	for _, d := range desired {
+		if !seen[d.Alias] {
+			merged = append(merged, d)
+		}
+	}
```
prezha: wouldn't this add back those currently "undesired" that were intended for removal?

ie, going though the existing list (that can include some from the hardcoded "desired" list), we skip any that should be removed, so they don't get flagged as "seen" (perhaps we should flag them above)
 
then here, we go through all hardcoded "desired" list again, and add any that were not "seen"

prezha: example test

```
func TestMergeConsolePluginProxy_RemovesDesiredAliases(t *testing.T) {
	ns := "openshift-storage"
	extra := consolev1.ConsolePluginProxy{
		Alias: "internalRgwS3",
		Endpoint: consolev1.ConsolePluginProxyEndpoint{
			Type: consolev1.ProxyTypeService,
			Service: &consolev1.ConsolePluginProxyServiceConfig{
				Name:      "rook-ceph-rgw-ocs-storagecluster-cephobjectstore",
				Namespace: ns,
				Port:      443,
			},
		},
		Authorization: consolev1.None,
	}

	existing := append(GetConsolePluginProxy(ns), extra)
	merged := MergeConsolePluginProxy(existing, ns, []string{"s3Vector"})

	for _, p := range merged {
		if p.Alias == "s3Vector" {
			t.Error("MergeConsolePluginProxy should remove entries matching removeAliases")
		}
	}

	desired := GetConsolePluginProxy(ns)
	if len(merged) != len(desired) {
		t.Errorf("expected %d proxy entries after removal, got %d", len(desired), len(merged))
	}
}
```

```
--- FAIL: TestMergeConsolePluginProxy_RemovesDesiredAliases (0.00s)
    /Users/Predrag.Rogic/dev/github.com/red-hat-storage/odf-operator/console/console_test.go:105: MergeConsolePluginProxy should remove entries matching removeAliases
    /Users/Predrag.Rogic/dev/github.com/red-hat-storage/odf-operator/console/console_test.go:111: expected 5 proxy entries after removal, got 6
```

ygalblum: I'm not sure this is a valid use case. 

This was the initial concern:
> The only concern I see is that if we ever rename or remove a "desired" proxy from the GetConsolePluginProxy list, then during an upgrade it will be treated as an "existing" manual entry by MergeConsolePluginProxy. And, it will never be removed from the ConsolePlugin CR, these "stale entries" could accumulate over time.

It refers to a case in which the alias is **not** in the `desired` list but because of this change, it will not be removed, leaving stale entries in the list.

However, in your test, you are trying to both remove and add the same alias in the same call.

prezha: thanks for the response!

i was thinking about the case where we receive one of the "desired" proxy in the `removeAliases` slice (passed to `MergeConsolePluginProxy` func) - that specific removal request would be ignored

if we do not expect that case to ever happen, then i think we should be fine

on the side note, in the example test above i used two _different_ plugins - adding one new ("internalRgwS3") and requesting removal of one from the "desired" list ("s3Vector")

SanjalKatiyar:
LGTM (we can merge once commits are squashed and `commitlint` is fixed).
@prezha @iamniting for the final approval.

console/console.go
SanjalKatiyar: if a same alias is present both under "desiredAliases" and "removeAliases" list, IMO it's fine to give the preference to desired list and let it override the removal list (existing behaviour in this PR)...

prezha: yes, that was basically what i wanted to confirm
thanks, @SanjalKatiyar

prezha:
/lgtm

thanks @ygalblum for fixing this!

ygalblum:
I've squashed the commits and shortened the commit tittle

SanjalKatiyar:
@iamniting can u plz check and tag ??

iamniting (changes requested):
@ygalblum Can you pls fix the commit msg problems?

ygalblum:
@iamniting I fixed the last issue I saw (I had to switch the Assisted-By and Sign-off lines). At this point I don't see any errors. But, there are blocked workflows. Please allow them so I can see if there are any additional issues

console/console.go
```diff
@@ -146,6 +146,47 @@ func GetConsolePluginProxy(serviceNamespace string) []consolev1.ConsolePluginPro
...
+
+	return merged
```
iamniting: @ygalblum IMO the current logic is a bit complicated, and we can simplify it!
```suggestion
	aliasesToRemove := make(map[string]bool, len(removeAliases))
	for _, alias := range removeAliases {
		aliasesToRemove[alias] = true
	}

	desired := GetConsolePluginProxy(serviceNamespace)

	desiredAliases := make(map[string]bool, len(desired))
	for _, proxy := range desired {
		desiredAliases[proxy.Alias] = true
	}

	merged := append([]consolev1.ConsolePluginProxy{}, desired...)

	for _, proxy := range existing {
		if _, remove := aliasesToRemove[proxy.Alias]; remove {
			continue
		}

		if _, found := desiredAliases[proxy.Alias]; found {
			continue
		}

		merged = append(merged, proxy)
	}

	return merged
```

iamniting (changes requested):

console/console.go
ygalblum: Thanks. I've updated the implementation based on your suggestion. One different, since the maps hold booleans no need for the second field assignment (which makes the code even simpler)

iamniting: Can you pls use the variable names as it is. It will make our life easier while looking at the code in the future.

iamniting (approved):
Thanks for Your PR @ygalblum!

PR #872, opened 2026-07-16
devicefinder: fix ConfigMap owner reference race on pod replacement

ezio-auditore:
Solves: https://redhat.atlassian.net/browse/DFBUGS-8878
Fix a race condition in devicefinder ConfigMap ownership during pod replacement (DaemonSet rolling updates, node drains, OOM kills).

Each node writes discovery results to a node-scoped ConfigMap (devicefinder-result-<node>). Previously, ownerReferences were set on the ConfigMap template before ctrl.CreateOrUpdate. On the update path, CreateOrUpdate loads the live object from the API server and overwrites the template — the mutate function only updated Data, so ownership was never transferred to the replacement pod.

This caused two problems:

Premature GC deletion — After Pod-A was deleted, the ConfigMap still referenced it. Once garbage collection processed the stale owner, the ConfigMap could be deleted while Pod-B was still running and writing to it.
Failed ownership reclaim on subsequent rotations — SetControllerReference returns AlreadyOwnedError when another pod is already the controller owner. Without clearing existing refs, the second rolling update would fail.
Changes
Store runtime.Scheme on DeviceDiscovery (required by SetControllerReference)
Refactor updateConfigMap to follow the CreateOrUpdate contract: template only sets Name/Namespace; mutate callback sets Data, Labels, clears existing ownerReferences, and calls ctrl.SetControllerReference for the current pod
Add unit test verifying a replacement pod reclaims ownership from a stale owner reference

### DIFF
--- a/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
+++ b/bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
@@ -7,7 +7,7 @@ metadata:
     categories: Storage
     console.openshift.io/plugins: '["odf-console"]'
     containerImage: quay.io/ocs-dev/odf-operator:latest
-    createdAt: "2026-06-26T07:30:49Z"
+    createdAt: "2026-07-20T09:34:53Z"
     description: OpenShift Data Foundation provides a common control plane for storage
       solutions on OpenShift Container Platform.
     features.operators.openshift.io/token-auth-aws: "true"
@@ -571,6 +571,7 @@ spec:
           - create
           - get
           - update
+          - delete
         serviceAccountName: ocs-devicefinder-sa
       - rules:
         - apiGroups:
--- a/services/devicefinder/devicefinder/discovery/configmap.go
+++ b/services/devicefinder/devicefinder/discovery/configmap.go
@@ -51,23 +51,26 @@ func (discovery *DeviceDiscovery) updateConfigMap() error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      configMapName,
 			Namespace: namespace,
-			Labels: map[string]string{
-				"app":  "devicefinder",
-				"node": nodeName,
-			},
-			OwnerReferences: []metav1.OwnerReference{
-				{
-					APIVersion: "v1",
-					Kind:       "Pod",
-					Name:       podName,
-					UID:        apimachinerytypes.UID(podUID),
-				},
-			},
+		},
+	}
+
+	ownerReference := []metav1.OwnerReference{
+		{
+			APIVersion: "v1",
+			Kind:       "Pod",
+			Name:       podName,
+			UID:        apimachinerytypes.UID(podUID),
 		},
 	}
 
 	_, err = ctrl.CreateOrUpdate(context.TODO(), discovery.kubeClient, configMap, func() error {
 		configMap.Data = configMapData
+		if configMap.Labels == nil {
+			configMap.Labels = make(map[string]string)
+		}
+		configMap.Labels["app"] = "devicefinder"
+		configMap.Labels["node"] = nodeName
+		configMap.OwnerReferences = ownerReference
 		return nil
 	})
 	if err != nil {
--- a/services/devicefinder/devicefinder/discovery/configmap_test.go
+++ b/services/devicefinder/devicefinder/discovery/configmap_test.go
@@ -1,0 +1,108 @@
+package discovery
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("updateConfigMap", func() {
+	const (
+		namespace  = "openshift-storage"
+		nodeName   = "worker-1"
+		oldPodName = "devicefinder-abc12"
+		oldPodUID  = "11111111-1111-1111-1111-111111111111"
+		newPodName = "devicefinder-def34"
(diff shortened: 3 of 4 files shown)


services/devicefinder/devicefinder/discovery/configmap.go
```diff
@@ -51,24 +51,25 @@ func (discovery *DeviceDiscovery) updateConfigMap() error {
...
+			Namespace: namespace,
+			UID:       apimachinerytypes.UID(podUID),
 		},
 	}
 
 	_, err = ctrl.CreateOrUpdate(context.TODO(), discovery.kubeClient, configMap, func() error {
+		configMap.SetOwnerReferences(nil)
 		configMap.Data = configMapData
-		return nil
+		configMap.Labels = map[string]string{
```
Madhu-1: we should not override the user added labels. lets have a unit test for it as well

ezio-auditore: fixed, also added unit-test

services/devicefinder/devicefinder/discovery/discovery.go
```diff
@@ -30,6 +30,7 @@ var supportedDeviceTypes = sets.NewString("mpath", "disk")
 // DeviceDiscovery instance
 type DeviceDiscovery struct {
 	kubeClient client.Client
+	scheme     *runtime.Scheme
```
rohan47: Is this required?

services/devicefinder/devicefinder/discovery/configmap.go
```diff
@@ -51,24 +52,26 @@ func (discovery *DeviceDiscovery) updateConfigMap() error {
...
-					APIVersion: "v1",
-					Kind:       "Pod",
-					Name:       podName,
-					UID:        apimachinerytypes.UID(podUID),
-				},
-			},
+		},
+	}
+
+	pod := &corev1.Pod{
```
rohan47: can we add the owner reference as it was done before but in CreateOrUpdate() ?
this will remove the need for using discovery.scheme bellow
```suggestion
	ownerReference := []metav1.OwnerReference{
				{
					APIVersion: "v1",
					Kind:       "Pod",
					Name:       podName,
					UID:        apimachinerytypes.UID(podUID),
				},
			},
```

services/devicefinder/devicefinder/discovery/configmap.go
```diff
@@ -51,24 +52,26 @@ func (discovery *DeviceDiscovery) updateConfigMap() error {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: namespace,
+			UID:       apimachinerytypes.UID(podUID),
 		},
 	}
 
 	_, err = ctrl.CreateOrUpdate(context.TODO(), discovery.kubeClient, configMap, func() error {
+		configMap.SetOwnerReferences(nil)
```
rohan47: Is this required?
We will replace the owner reference bellow

rohan47 (changes requested):

services/devicefinder/devicefinder/discovery/configmap.go
```diff
@@ -51,24 +52,26 @@ func (discovery *DeviceDiscovery) updateConfigMap() error {
...
 		configMap.Data = configMapData
-		return nil
+		if configMap.Labels == nil {
+			configMap.Labels = make(map[string]string)
+		}
+		configMap.Labels["app"] = "devicefinder"
+		configMap.Labels["node"] = nodeName
+		return controllerutil.SetOwnerReference(pod, configMap, discovery.scheme)
```
rohan47:
```suggestion
	    configMap.OwnerReferences = ownerReference
        return nil
```

ezio-auditore: Updated

services/devicefinder/devicefinder/discovery/configmap.go
ezio-auditore: updated

services/devicefinder/devicefinder/discovery/configmap.go
ezio-auditore: updated

services/devicefinder/devicefinder/discovery/discovery.go
ezio-auditore: not required anymore

ezio-auditore:
Tested for regression, no issues observed

services/devicefinder/devicefinder/discovery/configmap_test.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		discovery = &DeviceDiscovery{
+			kubeClient: fake.NewClientBuilder().WithScheme(scheme).WithObjects(staleCM).Build(),
+		}
+
+		Expect(os.Setenv("NODE_NAME", nodeName)).To(Succeed())
+		Expect(os.Setenv("POD_NAMESPACE", namespace)).To(Succeed())
+		Expect(os.Setenv("POD_NAME", newPodName)).To(Succeed())
+		Expect(os.Setenv("POD_UID", newPodUID)).To(Succeed())
```
Madhu-1: can we do something like this
```
t := GinkgoT()
		// Set the environment variable (scoped to this test)
		t.Setenv("MY_CUSTOM_VAR", "my_test_value")
```
we should not be doing os.Setenv i test files

ezio-auditore: ok, Updated

bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
```diff
@@ -571,6 +571,7 @@ spec:
           - create
           - get
           - update
+          - delete
```
Madhu-1, line 574: @ezio-auditore why do we need delete permission?

ezio-auditore: This is required for cases where stale configmap exists without a owneref. Faced this error while testing.
Adding delete permissions fixed it. 
`failed to update ConfigMap. Error: failed to create or update ConfigMap: configmaps "devicefinder-result-ip-10-0-19-193.ap-south-1.compute.internal" is forbidden: cannot set an ownerRef on a resource you can't delete:`

rohan47 (approved):

iamniting (approved):

PR #885, opened 2026-07-27
console: add NetworkPolicy for odf-console

weirdwiz:
## Summary

[RHSTOR-8737](https://redhat.atlassian.net/browse/RHSTOR-8737)

Add NetworkPolicies for the ODF console components:

- `odf-console`: allow ingress only from the `openshift-console` namespace on TCP port 9001 and deny all egress. ConsolePlugin proxy traffic originates from the OpenShift console pod, not from `odf-console`.
- `ux-backend-proxy`: allow ingress only from the `openshift-console` namespace on TCP port 8888.

Both policies are included in their parent kustomizations and generated into the OLM bundle through the existing bundle-generation path.

## Changes

- Add `config/console/odf-console-networkpolicy.yaml` to `config/console/kustomization.yaml`.
- Add `config/ux-backend/ux-backend-networkpolicy.yaml` to `config/ux-backend/kustomization.yaml`.
- Add the generated NetworkPolicy bundle manifests.
- Upgrade Operator SDK from v1.31.0 to v1.40.0, the first version used here that supports NetworkPolicy bundle resources, and update the project layout marker to Kubebuilder v4.
- Grant the controller create, update, and delete permissions for NetworkPolicies, as required for certification.

## Test plan

- [x] `make bundle`
- [x] `operator-sdk bundle validate bundle/odf-operator`
- [x] Verify `kustomize build config/manifests` renders both NetworkPolicies
- [ ] Deploy on a cluster with network policy enforcement and verify access is restricted to the expected namespace and ports


controllers/clusterversion_controller.go
```diff
@@ -60,6 +60,7 @@ type ClusterVersionReconciler struct {
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins,verbs=get;create;update
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consoleclidownloads,verbs=get;create;update
 //+kubebuilder:rbac:groups=console.openshift.io,resources=consolequickstarts,verbs=get;list;create;update;delete
+//+kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=create;update;delete
```
iamniting, line 64: Why do you need the permissions, when it is not being handled by the controller?

config/console/odf-console-networkpolicy.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: networking.k8s.io/v1
```
iamniting, line 1: pls rename this file to `odf-console-networkpolicy.yaml`

iamniting (changes requested):

bundle/odf-operator/manifests/odf-console_networking.k8s.io_v1_networkpolicy.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  egress: []
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-console
+    ports:
+    - port: 9001
+      protocol: TCP
```
SanjalKatiyar, line 16: is this level of granularity really needed ??

bundle/odf-operator/manifests/odf-console_networking.k8s.io_v1_networkpolicy.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: odf-console
+spec:
+  egress: []
```
SanjalKatiyar, line 8: it's tested, right ??

SanjalKatiyar, line 8: just FYI, for client-console pod (client-operator repo), we need to allow communication to external S3, so we need to be very careful with `egress` settings in that case...

SanjalKatiyar, line 8: just FYI, this will change in future, this pod will talk to external rgw, for now it's fine...

bipuladh (commented):
lgtm we prolly need the same for odf-client-console and mco-console.

bipuladh:
Also add policy for ux-backend-proxy

controllers/clusterversion_controller.go
weirdwiz: the operator does not reconcile these policies today. the rbac is added to avoid being flagged by ocp, as required by this guide: https://docs.google.com/document/d/1pChD6vsLmMd7PQPR9B7nn3wXBNLv_pP6Knhd-mDxk-Q/edit?tab=t.0

config/console/odf-console-networkpolicy.yaml
weirdwiz: done, renamed it to `odf-console-networkpolicy.yaml`.

bundle/odf-operator/manifests/odf-console_networking.k8s.io_v1_networkpolicy.yaml
weirdwiz: yes, this is intentional. the policy should only allow traffic from the `openshift-console` namespace to the plugin port `9001`. dropping either restriction would allow more traffic than needed.

bundle/odf-operator/manifests/odf-console_networking.k8s.io_v1_networkpolicy.yaml
weirdwiz: i’ve deployed the policy on an internal cluster, but i haven’t finished the client-side connectivity checks yet. i’ll verify the console flow and the allowed/denied traffic before marking testing complete.

weirdwiz: thanks for the heads-up. i’ll keep the client-console policy separate since its nginx proxy needs egress to external s3 endpoints.

weirdwiz: yep, agreed. deny-all egress matches the current odf-console behavior, but we’ll need to revisit it when the pod starts talking to external rgw.

weirdwiz:
@bipuladh added an ingress-only policy for `ux-backend-server` under the ux-backend config. it allows traffic from `openshift-console` to port `8888` and leaves egress unchanged since the backend and oauth proxy need outbound access.

agreed that odf-client-console and mco-console need policies too. they’re deployed by different operators, so the corresponding changes will be made in those operator repos.

controllers/clusterversion_controller.go
iamniting: Can you pls add it as a separate commit? We should also add a comment in the Go file. So that we don't remove it accidentally

config/console/odf-console-networkpolicy.yaml
iamniting: I believe you made the changes locally, but forgot to push them.

controllers/clusterversion_controller.go
weirdwiz: Done the RBAC change is now in a separate commit

config/console/odf-console-networkpolicy.yaml
weirdwiz: Pushed now the source file is named `odf-console-networkpolicy.yaml` and the kustomization references it.

iamniting (changes requested):
Changes looks good to me, Can you pls divide the first commit into 3 commits, 
first one for the manual changes
second one for the bundle and other generated changes
third one for the catalog generated changes.

bundle/odf-operator/manifests/ux-backend-proxy_networking.k8s.io_v1_networkpolicy.yaml
```diff
@@ -0,0 +1,18 @@
...
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-console
+    ports:
+    - port: 8888
+      protocol: TCP
+  podSelector:
+    matchLabels:
+      app: ux-backend-server
+  policyTypes:
+  - Ingress
```
SanjalKatiyar, line 18: are we skipping the egress because pod needs to talk to the api-server, or because this is a server pod that may need to talk to different backends in the future, and we don't want to restrict that ??

weirdwiz: egress is intentionally left open since ux-backend-server and oauth-proxy need outbound access to the kubernetes api server and odf services. this is consistent with the ingress-only policies in ocs-client-operator and odf-multicluster-orchestrator.

SanjalKatiyar:
LGTM.

iamniting (approved):

PR #895, opened 2026-08-07
controllers: keep relevant CRDs in cache with client-go v0.36

iamniting:
Returning nil for non-relevant CRDs from the transformer aborted cache
replacement, causing even relevant CRDs to be removed from the cache.
Keep name-only stubs for non-relevant CRDs instead, and ignore those
stubs in the OperatorScaler CRD watch predicate.

This restores CRD presence in the informer cache so OperatorScaler can
get/list CRDs in the configmap, register dynamic watches on the
corresponding CRs, scale dependent operator CSV deployments, and
export storage-system metrics.


leelavg:
```
REGRESSION SUMMARY: client-go v0.36 RealFIFO Transform Nil Handling

Root Cause:
controller-runtime v0.19 (matching k8s v0.36) switched default informer cache from DeltaFIFO to RealFIFO with AtomicEvents enabled. RealFIFO v0.36 changed nil semantics:
v0.35 filtered nil returns implicitly, v0.36 preserves them explicitly in batch Replace operations.

Impact:
Transform functions returning nil to skip items now drop those items entirely from cache. CRDs disappeared from OperatorScaler cache, breaking dynamic watch registration
and scaling.

Key Commits & PR:
- PR: kubernetes/kubernetes #135462 "Add atomic replace in client-go" (merged Jan 14, 2026)
- client-go commits:
- 0a8284b969181c300604fe6d114dd2a6c8d1e475 - Plumb AtomicFIFO options into controller/store construction
- aad3a6de8992aed21c6878660d2d7722375edbac - Add support for ReplacedAll delta handling
- Released: client-go v0.36.0 (April 22, 2026)

Semantic Change in RealFIFO.Replace():
- v0.35: Transform results handled per-item, nil implicit
- v0.36: Transform results (including nil) explicitly stored in array without filtering in addReplaceToItemsLocked() method

Fix Applied (commit 0753dec):
Return name-only stub CRD with only ObjectMeta.Name instead of nil. Filter stubs in OperatorScaler's CRD watch predicate by checking crd.Spec.Group != "". Stubs pass
through Replace, relevant CRDs stay cached, stubs ignored.
```

kubernetes/kubernetes#135462
/lgtm

leelavg:
is the *crd change not working, code looks correct though.

iamniting:
> is the *crd change not working, code looks correct though.

It is working the issue is with the ocp ci

PR #899, opened 2026-08-11
Pin GitHub actions by SHA

AnikaYadav1:
Pin all third-party actions to immutable commit SHA references.


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
+version: 2
```
iamniting: @AnikaYadav1 Can you try this on your remote first, Also lets run it on main branch only and every monday.

iamniting (changes requested):
Also I would like to keep the dependabot a seperate PR.

iamniting (approved):

PR #938, opened 2026-09-18
fix nginx worker_processes OOM on high-CPU nodes

NIKHITHAVADDEMPUDI:
https://redhat.atlassian.net/browse/DFBUGS-10944

**Problem:** odf-console is OOMKilled on high-CPU nodes because nginx uses worker_processes auto and spawns roughly one worker per CPU.

**Changes:**

- Operator Deployment/CSV now sets NGINX_WORKER_PROCESSES=8 on the odf-operator pod by default.
- On reconcile, odf-operator calls GenerateNginxConf().
- GenerateNginxConf() reads NGINX_WORKER_PROCESSES via GetNginxWorkerProcesses().
- If the env is unset/invalid → use 8; if set to a positive integer or auto → use that value.
- The generated nginx config now contains worker_processes <value>; instead of worker_processes auto;.
- That config is written into the odf-console nginx ConfigMap by the reconciler.
- The odf-console pod mounts the ConfigMap, so nginx starts with a fixed small worker count (default 8).
- Manual edits to the ConfigMap are still overwritten on the next reconcile; change the operator env if a different value is needed.
- Unit tests cover the default (8), env override, and invalid fallback behavior.


bundle/odf-operator/manifests/odf-operator.clusterserviceversion.yaml
```diff
@@ -442,6 +442,8 @@ spec:
                 - name: ONBOARDING_TOKEN_LIFETIME
                 - name: UX_BACKEND_PORT
                 - name: TLS_ENABLED
+                - name: NGINX_WORKER_PROCESSES
+                  value: "8"
```
SanjalKatiyar, line 446: I think it's fine if we don't add it to the CSV...
controller can look for the flag and if not found can default to `8`.

We can document updating Subscription in case this value needs to be changed at runtime...

console/nginx_conf.go
```diff
@@ -18,16 +18,30 @@ package console
...
+const (
+	// DefaultNginxWorkerProcesses is used when NGINX_WORKER_PROCESSES is unset
+	// or invalid. A fixed value avoids OOM/FD exhaustion on high-CPU nodes
+	// where "auto" would spawn one worker per CPU.
+	DefaultNginxWorkerProcesses = "8"
+
+	// NginxWorkerProcessesEnvVar overrides DefaultNginxWorkerProcesses when set
+	// on the operator pod (e.g. via Subscription config.env). Accepted values
+	// are a positive integer or "auto".
+	NginxWorkerProcessesEnvVar = "NGINX_WORKER_PROCESSES"
```
SanjalKatiyar, line 37:
```suggestion
	NginxWorkerProcessesEnvVar = "CONSOLE_NGINX_WORKER_PROCESSES"
```

SanjalKatiyar (changes requested):
plz test the changes...

NIKHITHAVADDEMPUDI:
Screenshot 2026-09-22 at 1 30 12 PM

https://github.com/user-attachments/assets/c94662f8-a8e0-4db9-8d72-5116f96dd09d


https://github.com/user-attachments/assets/f18242f0-17c1-41e2-a20a-eb3e94b1f1f4

SanjalKatiyar:
LGTM, but @NIKHITHAVADDEMPUDI plz fix commit formatting (check CI error).

cc @iamniting for approval.

iamniting (changes requested):
@NIKHITHAVADDEMPUDI Can you pls fix the DCO?

NIKHITHAVADDEMPUDI:
@iamniting I had formatted the commit message, could you please provide approval for this PR

iamniting (approved):

